PR #6, opened 2019-06-27
Threaded context manager

shannynalayna:
Adds the implementation for the threaded Context Manager, including a ContextManagerCore which holds an internal context manager which is used based on the messages received. Also adds a ContextManagerJoinHandle which is used to start the ContextManager thread and to create ContextManagers which hold a sender to send messages to the ContextManagerCore.

Changes the references to the thread-safe Context Manager in the sync module used in the static execution adapter to the threaded version. Also changes the ContextLifecycle trait to be used by the Context Manager thread handler so the scheduler is using the ContextManagerJoinHandle and threaded ContextManager.

### DIFF
--- a/libtransact/src/context/error.rs
+++ b/libtransact/src/context/error.rs
@@ -15,7 +15,11 @@
  * -----------------------------------------------------------------------------
  */
 use std::error::Error;
+use std::sync::mpsc::{RecvError, SendError};
 
+use crate::context::manager::thread::{
+    ContextManagerCoreError, ContextOperationMessage, ContextOperationResponse,
+};
 use crate::protocol::receipt::TransactionReceiptBuilderError;
 use crate::state::error::StateReadError;
 
@@ -24,6 +28,7 @@ pub enum ContextManagerError {
     MissingContextError(String),
     TransactionReceiptBuilderError(TransactionReceiptBuilderError),
     StateReadError(StateReadError),
+    InternalError(Box<dyn Error>),
 }
 
 impl Error for ContextManagerError {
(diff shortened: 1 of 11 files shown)


libtransact/src/context/error.rs
```diff
@@ -71,3 +73,71 @@ impl From<StateReadError> for ContextManagerError {
         ContextManagerError::StateReadError(err)
     }
 }
+
+#[derive(Debug)]
+pub enum ContextManagerCoreError {
```
ltseeley: I wonder if it would make more sense for this to be in the `thread` sub-module, since that's the module that "owns" this error type?

libtransact/src/context/manager/thread/mod.rs
```diff
@@ -272,3 +272,163 @@ impl ContextManager {
...
+    fn make_event() -> Event {
+        EventBuilder::new()
+            .with_event_type(EVENT_TYPE1.to_string())
+            .with_attributes(vec![
+                (ATTR1.0.to_string(), ATTR1.1.to_string()),
+                (ATTR2.0.to_string(), ATTR2.1.to_string()),
+            ])
+            .with_data(BYTES1.to_vec())
+            .build()
+            .unwrap()
```
ltseeley: Generally `expect` is more useful for tests, because it's easier to identify what failed (unwrap doesn't provide any useful information).

libtransact/src/context/manager/thread/mod.rs
```diff
@@ -272,3 +272,163 @@ impl ContextManager {
...
+    /// Context Manager thread.
+    #[test]
+    fn test_context_lifecycle() {
+        let state = HashMapState::new();
+        let (join_handle, context_manager) = ContextManagerJoinHandle::new(Box::new(state));
+
+        let state_id = HashMapState::state_id(&HashMap::new());
+
+        let context_id = context_manager.create_context(&[], &state_id);
+        assert!(context_id.is_ok());
```
ltseeley: Rather than calling `assert!(_.is_ok())`, you can just do `_.expect("failed to ...")`

PR #8, opened 2019-06-28
Initial implementation of Command transaction family

shannynalayna:
Implements the Command transaction family through the workload module. Adds Commands defined by the Command Family RFC that were not previously implemented. Also adds tests for these commands via the static execution adapter.

### DIFF
--- a/libtransact/protos/command.proto
+++ b/libtransact/protos/command.proto
@@ -1,0 +1,121 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+syntax = "proto3";
+
+// A transaction payload for the Command Family
+message CommandPayload {
+    // The list of commands to execute
+    repeated Command commands = 1;
+}
+
+// A key-value pair where the value is a bytes field.
+message BytesEntry {
+    // The key, for instance a state address or a string value, depending on
+    // the target of the entry.
+    string key = 1;
+    // The byte value to be stored
+    bytes value = 2;
+}
+
+// The list of state keys to be retrieved via a GET_STATE command.
+message GetState {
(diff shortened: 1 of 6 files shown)


libtransact/src/workload/command.rs
```diff
@@ -159,13 +197,13 @@ impl std::str::FromStr for Command {
         let command = command_parts.next().map(|s| s.to_lowercase());
         match command {
             Some(ref set) if set == "set" => {
-                let address = command_parts
```
peterschwarz: All of this parsing should now be replaced with just a `protobuf::parse_from_bytes`

libtransact/src/workload/command.rs
```diff
@@ -75,24 +96,41 @@ impl TransactionHandler for CommandTransactionHandler {
...
-    Get { address: String },
-    Fail { error_msg: String },
-}
-
 impl std::fmt::Display for Command {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
-            Command::Set {
-                ref address,
+            Command::SetState {
```
peterschwarz: This display implementation is just for rendering these commands for the tests also replaced by the protobuf classes.

libtransact/protos/command.proto
```diff
@@ -0,0 +1,121 @@
...
+message AddReceiptData {
+    // a non-empty byte array
+    bytes receipt_data = 1;
+}
+
+// The duration and method of sleep to apply during a SLEEP command.
+message Sleep {
+    enum SleepType {
+        // This will wait, releasing cpu utilization to other processes
+        WAIT = 0;
```
ltseeley: Maybe "IDLE" and "BUSY" would be better?

libtransact/src/protocol/command.rs
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------------
+ */
+
+/// A command to be executed
+#[derive(Debug)]
+pub enum Command {
+    SetState {
+        state_keys: String,
```
ltseeley: Should this be `state_key`? Same below.

libtransact/src/protocol/command.rs
```diff
@@ -0,0 +1,55 @@
...
+    },
+    DeleteState {
+        state_keys: String,
+    },
+    GetState {
+        state_keys: String,
+    },
+    AddEvent {
+        event_type: String,
+        attributes: Vec<(String, String)>,
```
ltseeley: Should this be `Vec<(String, Vec<u8>)>`?

libtransact/protos/command.proto
shannynalayna: This is the SleepType that was defined in the Command family RFC.

PR #9, opened 2019-07-03
Refactor transaction protocol using protocol macros

RyanLassigBanks:
Adds macros for generating the following traits and uses them with `Batch` and `TransactionHeader`

- FromProto
- FromNative
- FromBytes
- IntoProto
- IntoNative
- IntoBytes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,2 @@
 [workspace]
-members = ["libtransact"]
+members = ["libtransact", "transact_derive"]
--- a/libtransact/src/protocol/batch.rs
+++ b/libtransact/src/protocol/batch.rs
@@ -31,6 +31,7 @@ use crate::protos::{
     FromBytes, FromNative, FromProto, IntoBytes, IntoNative, IntoProto, ProtoConversionError,
 };
 use crate::signing;
+use transact_derive::FromProtoImpl;
 
 use super::transaction::Transaction;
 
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["libtransact"]
+members = ["libtransact", "transact_derive"]
```
peterschwarz: Commit message is wrong - "Add transact-derive"

Also, add some context in the commit message describing what the library includes, as of this commit.

transact_derive/src/protos.rs
```diff
@@ -0,0 +1,624 @@
...
+        if meta.name() != Ident::new("proto_enum", Span::call_site()) {
+            continue;
+        }
+
+        let meta_list = if let Meta::List(l) = meta.clone() {
+            l.nested
+        } else {
+            return Err(SynError::new_spanned(
+                meta.into_token_stream(),
+                "Malformed Meta",
```
peterschwarz: Any way there could be more context with these "Malformed Meta" message?

libtransact/src/protocol/batch.rs
```diff
@@ -106,10 +107,14 @@ impl IntoBytes for BatchHeader {
 impl IntoProto<protos::batch::BatchHeader> for BatchHeader {}
 impl IntoNative<BatchHeader> for protos::batch::BatchHeader {}
 
-#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+#[derive(FromProtoImpl, Debug, Clone, Eq, Hash, PartialEq)]
+#[proto_type = "protos::batch::Batch"]
 pub struct Batch {
     header: Vec<u8>,
+    #[from_proto_impl(to_string)]
```
peterschwarz: These seem like they should be detectable.

PR #10, opened 2019-07-08
fixing the authors of this crate

dhuseby:
Signed-off-by: Dave Huseby <dhuseby@linuxfoundation.org>

### DIFF
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "transact"
 version = "0.1.4"
-authors = ["Bitwise IO, Inc.", "Cargill Incorporated"]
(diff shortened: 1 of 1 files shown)


dhuseby:
Once this lands, please update the crates.io page using cargo:
https://doc.rust-lang.org/cargo/reference/publishing.html

dhuseby:
@vaporos the Apache License file lists the copyright holders. The authors is different.

dhuseby:
@vaporos The right thing to do is to list the copyright as the Transact Authors and then have an AUTHORS file with the copyright holders. See [here](https://opensource.google.com/docs/releasing/authors/) for an independent take on this from Google's open source page.

PR #11, opened 2019-07-29
Native/proto conversion implementations

ltseeley:
Implements the various native/protobuf conversion traits for the Transaction, Batch, and BatchPair structs. This makes it easier to deal with batches when using the library by cutting out conversion steps (e.g. can go from bytes -> BatchPair with a single method call).

### DIFF
--- a/libtransact/src/protocol/batch.rs
+++ b/libtransact/src/protocol/batch.rs
@@ -141,6 +141,61 @@ impl Batch {
     }
 }
 
+impl FromProto<protos::batch::Batch> for Batch {
+    fn from_proto(batch: protos::batch::Batch) -> Result<Self, ProtoConversionError> {
+        Ok(Batch {
+            header: batch.header,
+            header_signature: batch.header_signature,
+            transactions: batch
+                .transactions
+                .into_iter()
+                .map(|txn| txn.into_native())
+                .collect::<Result<Vec<_>, _>>()?,
+            trace: batch.trace,
+        })
+    }
+}
+
+impl FromNative<Batch> for protos::batch::Batch {
+    fn from_native(batch: Batch) -> Result<Self, ProtoConversionError> {
+        let mut proto_batch = protos::batch::Batch::new();
+        proto_batch.set_header(batch.header);
+        proto_batch.set_header_signature(batch.header_signature);
+        proto_batch.set_transactions(
+            batch
+                .transactions
+                .into_iter()
+                .map(|txn| txn.into_proto())
+                .collect::<Result<Vec<_>, _>>()?
+                .into(),
+        );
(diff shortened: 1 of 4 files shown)


libtransact/src/protocol/batch.rs
```diff
@@ -141,6 +141,61 @@ impl Batch {
...
+        );
+        proto_batch.set_trace(batch.trace);
+        Ok(proto_batch)
+    }
+}
+
+impl FromBytes<Batch> for Batch {
+    fn from_bytes(bytes: &[u8]) -> Result<Batch, ProtoConversionError> {
+        let proto: protos::batch::Batch = protobuf::parse_from_bytes(bytes).map_err(|_| {
+            ProtoConversionError::SerializationError("Unable to get Batch from bytes".to_string())
```
vaporos: This should probably include the error string from the causing error. "Unable to get Batch from bytes: {}"

libtransact/src/protos.rs
```diff
@@ -44,6 +47,9 @@ impl StdError for ProtoConversionError {
 impl std::fmt::Display for ProtoConversionError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match *self {
+            ProtoConversionError::DeserializationError(ref s) => {
+                write!(f, "DeserializationError: {}", s)
```
peterschwarz: These are effectively the equivalent of the `Debug` output.  These should be more of a human-readable display string.

libtransact/src/protocol/batch.rs
ltseeley: Added commit to update the error messages

libtransact/src/protos.rs
ltseeley: Added commit to update these error messages

libtransact/src/protos.rs
```diff
@@ -48,12 +48,14 @@ impl std::fmt::Display for ProtoConversionError {
...
             ProtoConversionError::DeserializationError(ref s) => {
-                write!(f, "DeserializationError: {}", s)
+                write!(f, "unable to deserialize during protobuf conversion: {}", s)
             }
             ProtoConversionError::SerializationError(ref s) => {
-                write!(f, "SerializationError: {}", s)
+                write!(f, "unable to serialize during protobuf conversion: {}", s)
+            }
+            ProtoConversionError::InvalidTypeError(ref s) => {
+                write!(f, "invalid type encountered protobuf conversion: {}", s)
```
peterschwarz: missing "during" - should be "...encountered during protobuf..."

ltseeley: fixed

PR #17, opened 2019-09-20
[TX-54] convert btw state and receipt StateChange

swetharepakula:
- since transaction receipt state change enum cannot be used as is
 when applying updates to state, have a method to convert between the
 two

Signed-off-by: Swetha Repakula <srepaku@us.ibm.com>

### DIFF
--- a/libtransact/src/protocol/receipt.rs
+++ b/libtransact/src/protocol/receipt.rs
@@ -22,6 +22,7 @@ use crate::protos;
 use crate::protos::{
     FromBytes, FromNative, FromProto, IntoBytes, IntoNative, IntoProto, ProtoConversionError,
 };
+use crate::state;
 use std::error::Error as StdError;
 use std::fmt;
 
@@ -45,6 +46,16 @@ impl StateChange {
             key == k
         } else {
             false
+        }
+    }
+}
+
+impl Into<state::StateChange> for StateChange {
+    /// Converts this modules StateChange into state::StateChange enum
+    fn into(self) -> state::StateChange {
+        match self {
+            StateChange::Set { key, value } => state::StateChange::Set { key, value },
+            StateChange::Delete { key } => state::StateChange::Delete { key },
         }
     }
 }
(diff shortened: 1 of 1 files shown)


swetharepakula:
I think some of the naming could be improved but could not think of a function that would be clearer.

vaporos:
This will need to be rebased once #19 merges.

libtransact/src/protocol/receipt.rs
```diff
@@ -49,6 +50,16 @@ impl StateChange {
     }
 }
 
+impl Into<state::StateChange> for StateChange {
```
RyanLassigBanks: Since `state` and `protocol` live in the same crate, you'd actually want to implement `From` on `state::StateChange` which will implicitly implement `Into` on `receipt::StateChange`

https://doc.rust-lang.org/std/convert/trait.Into.html

vaporos: Can you do that within the receipt module to avoid importing receipt into the state module?

swetharepakula: I implemented Into instead of From because of this conversation on rocket chat: https://chat.hyperledger.org/channel/transact?msg=aEaCj3J6R5w26m6BJ.

The reasoning was that receipt already does conversions so it should support the conversion and not the other way around, and we only need the conversion one way. We wouldn't need to go from state to receipt.

Please let me know if that is no longer the decision, I can definitely switch to From. As you said we would get the Into for free.

RyanLassigBanks: You can since they're in the same crate. Personally, I think `From`s should be in the same file as the struct definition, but it's more than valid.

vaporos:
Would be good to rewrite "btw" to "between" in the commit message, and capitalize the first letter of the first line.

libtransact/src/protocol/receipt.rs
RyanLassigBanks: However, if the intention is to have protocol be it's own crate eventually then I think `Into` is fine.

swetharepakula:
Fixed the commit message

libtransact/src/protocol/receipt.rs
swetharepakula: I didn't realize you could implement From in the receipt module. Just to clarify then, which trait would we like?

RyanLassigBanks: It should be

```
impl From<StateChange> for receipt::StateChange {
...
}
```

swetharepakula: Great, updated to implement the `From` trait. Thanks!

RyanLassigBanks: So I just had a spirited discussion with @vaporos, and we came to the conclusion that `Into` would actually be the correct trait since state shouldn't be aware of protocol. I'm really sorry for the inconvenience/ confusion.

swetharepakula: Okay, no problem. Not a difficult change to make :). Updated the PR.

vaporos: For context - the reason we think Into is the right thing here, is that From would add to the state::State API, which we were explicitly trying to keep "clean" of receipt. Even if the From is implemented in the receipt module, the API of state::State changes. Thus the reasoning behind keeping it only Into.

PR #20, opened 2019-09-30
Extract Sawtooth XO Test to an Example

swetharepakula:
- the example uses the XO transaction family to play a game of tic tac
 toe.
 - the example shows how to  use trasact end to end with multiple batches
 with state changes to the database in between

Signed-off-by: Swetha Repakula <srepaku@us.ibm.com>

### DIFF
--- a/examples/simple_xo/Cargo.toml
+++ b/examples/simple_xo/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "simple-xo"
+version = "0.1.0"
+authors = ["IBM Corp.", "Cargill Incorporated"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+transact = { version = "0.1.6", features = ["sawtooth-compat"] }
+sawtooth-xo = "0.3"
+hex = "0.3"
+sha2 = "0.8"
--- a/examples/simple_xo/src/main.rs
+++ b/examples/simple_xo/src/main.rs
@@ -1,0 +1,261 @@
+// Copyright 2019 IBM Corp.
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use sawtooth_xo::handler::XoTransactionHandler;
+use transact::context::manager::sync::ContextManager;
+use transact::database::{btree::BTreeDatabase, Database};
+use transact::execution::{adapter::static_adapter::StaticExecutionAdapter, executor::Executor};
+use transact::protocol::receipt::TransactionReceipt;
+use transact::protocol::{
+    batch::{BatchBuilder, BatchPair},
+    receipt::StateChange,
+    transaction::{HashMethod, TransactionBuilder},
+};
+use transact::sawtooth::SawtoothToTransactHandlerAdapter;
+use transact::scheduler::{
+    serial::SerialScheduler, BatchExecutionResult, Scheduler, TransactionExecutionResult,
+};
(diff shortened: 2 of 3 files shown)


examples/simple_xo/src/main.rs
```diff
@@ -0,0 +1,246 @@
+use sawtooth_xo::handler::XoTransactionHandler;
```
vaporos: Please add copyright headers to all the files (following the other files in the repo). Don't just blindly copy the 'Copyright' line though, make sure it's accurate. If you feel this is sufficiently derived from other code, include multiple 'Copyright' lines (the one(s) from the other code plus one for (presumably) IBM).

examples/simple_xo/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "simple_xo"
+version = "0.1.0"
+authors = ["Swetha Repakula <srepaku@us.ibm.com>"]
```
vaporos: Make this match the copyright holder(s). This is an active topic of discussion, as to what we should put here, but that's the approach used in the other Cargo.toml file currently.

examples/simple_xo/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "simple_xo"
```
vaporos: Both _ and - are valid here, but we've been preferring '-' based on what other "good" projects chose to do.

examples/simple_xo/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+fn print_current_state(value: &Vec<u8>) -> &str {
+    let val = match str::from_utf8(&value) {
+        Ok(v) => v,
+        Err(e) => panic!("Invalid UTF-8 sequence: {}", e),
+    };
+
+    let split: Vec<&str> = val.split(",").collect();
```
vaporos: A future enhancement (maybe after this PR?) might be create a Board struct and implement Display for it, and then do Board::from(&str) and println!("{}", board). Same with Game actually.

examples/simple_xo/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+use transact::signing::{hash::HashSigner, Signer};
+use transact::state::merkle::{self, MerkleRadixTree, MerkleState};
+use transact::state::StateChange as ChangeSet;
+use transact::state::Write;
+
+use hex;
+use sha2::{Digest, Sha512};
+use std::io;
+use std::str;
+use std::sync::{Arc, Mutex};
```
vaporos: We should be able to write this example without Arc/Mutex. But, I am also comfortable with this version going in and then iterating on it as an overall team.

examples/simple_xo/Cargo.toml
swetharepakula: Should I do the same with the directory name?

examples/simple_xo/src/main.rs
swetharepakula: I like that! I will work on that

examples/simple_xo/Cargo.toml
vaporos: No, everywhere else _ is appropriate.

vaporos:
Tested this. Very cool to have this be an interactive example.

Currently, it is very easy to cause an error which will make the demo exit. We may want to put in a feature in a subsequent PR that prevents invalid transactions from being fatal.

vaporos:
Needs updates so that the build passes (currently failing on clippy errors).

examples/simple_xo/src/main.rs
peterschwarz: I can see why this was done (based on the tests), but it definitely works without.

swetharepakula:
I am not sure how to fix this error: https://build.sawtooth.me/job/Transact-Hyperledger/job/transact/job/PR-20/6/execution/node/23/log/

rberg2:
> I am not sure how to fix this error: https://build.sawtooth.me/job/Transact-Hyperledger/job/transact/job/PR-20/6/execution/node/23/log/

Hello, this just means that a maintainer will need to start the build for you. I have started the build now. I apologize for not noticing and starting this sooner. Feel free to ping my on rocket chat if this happens again.

swetharepakula:
Thanks!

examples/simple_xo/src/main.rs
arsulegai: Because there's one instance of executor and the execution happen serially?

swetharepakula:
Hi, what are the remaining things left for this PR to be merged in?

PR #23, opened 2019-10-08
Add whitelist check to Jenkinsfile

rbuysse:
Signed-off-by: Ryan Beck-Buysse <rbuysse@bitwise.io>

### DIFF
--- a/bin/whitelist
+++ b/bin/whitelist
@@ -1,0 +1,35 @@
+#!/bin/bash
+#
+# Copyright 2017 Intel Corporation
+#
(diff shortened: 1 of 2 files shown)


bin/whitelist
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+#
+# Copyright 2017 Intel Corporation
```
ltseeley: Should the copyright be updated since this is a new file?

rbuysse: It's a copy of the whitelist file from sawtooth-core so I believe the original copyright is still correct.

PR #24, opened 2019-10-09
Added secp256k1 signing from ursa

mikelodder7:
Adds secp256k1 signatures from Ursa as a signer.


ltseeley:
Looks like this needs to be run through rustfmt

PR #26, opened 2019-11-18
Capture invalid results in TransactionReceipts

ltseeley:
Add fields to the `TransactionReceipt` protobuf that allow storing the
results of invalid transactions as receipts, not just valid
transactions. This will allow applications to store the results of all
transactions, whether valid or not, alongside each other.

This commit updates the transaction receipt and scheduler Rust APIs.


libtransact/src/protocol/receipt.rs
```diff
@@ -420,37 +461,37 @@ impl std::fmt::Display for TransactionReceiptBuilderError {
 }
 
 #[derive(Default, Clone)]
-pub struct TransactionReceiptBuilder {
+pub struct ValidTransactionReceiptBuilder {
     pub state_changes: Vec<StateChange>,
```
peterschwarz: Not sure I understand why these are all public

libtransact/src/protocol/receipt.rs
```diff
@@ -420,37 +461,37 @@ impl std::fmt::Display for TransactionReceiptBuilderError {
 }
 
 #[derive(Default, Clone)]
-pub struct TransactionReceiptBuilder {
+pub struct ValidTransactionReceiptBuilder {
```
peterschwarz: Shouldn't there be an analogous `InvalidTransactionReceiptBuilder`?  Or a single builder that would do something like:

```
TransactionReceiptBuilder::new()
    .valid()
    .with_state_changes(...)
   ...
```
and 
```
TransactionReceiptBuilder::new()
    .invalid()
    .with_message("Some message".to_string())
    ...
```

libtransact/src/protocol/receipt.rs
ltseeley: Updated

libtransact/src/protocol/receipt.rs
ltseeley: Updated

libtransact/protos/transaction_receipt.proto
```diff
@@ -18,15 +18,34 @@ syntax = "proto3";
 import "events.proto";
 
 message TransactionReceipt {
+  enum Result {
+    VALID = 0;
```
agunde406: Do we want the default to be VALID?  We usally have an enum value like "UNSET_RESULT = 0" for protobuf enums

libtransact/src/protocol/receipt.rs
```diff
@@ -164,62 +164,103 @@ impl IntoBytes for StateChange {
...
-                .events
-                .into_iter()
-                .map(protos::events::Event::from_native)
-                .collect::<Result<Vec<protos::events::Event>, ProtoConversionError>>()?,
-        ));
-        proto_transaction_receipt
-            .set_data(protobuf::RepeatedField::from_vec(transaction_receipt.data));
         proto_transaction_receipt.set_transaction_id(transaction_receipt.transaction_id);
+
+        match transaction_receipt.transaction_result {
```
agunde406: I don't see the result enum getting set

libtransact/protos/transaction_receipt.proto
ltseeley: Updated

libtransact/src/protocol/receipt.rs
ltseeley: Added

libtransact/protos/transaction_receipt.proto
ltseeley: Actually, I'm not sure how to handle this, since it's breaks compatibility with Sawtooth 1.0, and therefore causes the `protocol::receipt::tests::transaction_receipt_sawtooth10_compatibility` test to fail.

ltseeley: Updated to interpret an unset result as VALID when parsing a protobuf as a native Rust struct; this maintains compatibility with Sawtooth 1.x

PR #31, opened 2019-12-23
Add Addresser example

shannynalayna:
Adds an `address_generator` module in /examples to hold an example
program which takes in a prefix and a natural key, up to three strings,
to calculate a radix address.

Signed-off-by: Shannyn Telander <telander@bitwise.io>

### DIFF
--- a/examples/address_generator/Cargo.toml
+++ b/examples/address_generator/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "address-generator"
+version = "0.1.0"
+authors = ["Cargill Incorporated"]
+edition = "2018"
+
+[dependencies]
+clap = "2"
+
+[dependencies.transact]
+path = "../../libtransact"
+features = [
+    "contract-address-double-key-hash",
+    "contract-address-key-hash",
+    "contract-address-triple-key-hash",
+]
--- a/examples/address_generator/src/main.rs
+++ b/examples/address_generator/src/main.rs
@@ -1,0 +1,127 @@
+//
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::{App, Arg};
+
+use transact::contract::address::{
+    double_key_hash::DoubleKeyHashAddresser, key_hash::KeyHashAddresser,
+    triple_key_hash::TripleKeyHashAddresser, Addresser,
+};
+
+fn main() -> Result<(), String> {
+    let matches = App::new(clap::crate_name!())
+        .version(clap::crate_version!())
+        .author(clap::crate_authors!())
+        .about(clap::crate_description!())
+        .arg(
+            Arg::with_name("prefix")
+                .required(true)
+                .takes_value(true)
+                .help("Prefix of the radix address")
+                .index(1),
+        )
+        .arg(
+            Arg::with_name("key")
+                .required(true)
+                .multiple(true)
+                .takes_value(true)
+                .max_values(3)
+                .index(2)
+                .help("Natural key used to compute a radix address"),
+        )
+        .arg(
+            Arg::with_name("length")
+                .long("length")
+                .required(false)
+                .max_values(2)
+                .multiple(true)
+                .use_delimiter(true)
+                .help("Length of hash used to construct a radix address"),
+        )
+        .get_matches();
+
+    // The prefix is used to construct the Addresser.
+    let prefix = matches
+        .value_of("prefix")
+        .expect("Unable to get argument: prefix")
+        .to_string();
+
+    // Key(s) to be hashed and used to construct the radix address.
+    let keys: Vec<_> = matches
+        .values_of("key")
+        .expect("Unable to get argument: key")
+        .collect();
+
+    // Length(s) of the hashes used to construct the radix address.
+    let lengths = match matches.values_of("length") {
+        Some(vals) => vals
+            .map(|l| {
+                l.parse()
+                    .map_err(|_| "Argument must be an integer: length".to_string())
+            })
+            .collect::<Result<Vec<usize>, String>>()?,
+        _ => vec![],
+    };
+
+    // The Addresser implementation to be constructed depends on the amount of `keys` supplied.
+    // Once the appropriate Addresser implementation is constructed, the radix address will be
+    // calculated and printed out.
+    match keys.len() {
+        1 => {
+            let addresser = KeyHashAddresser::new(prefix);
+            let addr = addresser
+                .compute(&keys[0].to_string())
+                .map_err(|e| format!("Unable to compute address: {}", e))?;
+            println!("{}", addr);
+        }
+        2 => {
+            let addresser = match lengths.len() {
+                0 => DoubleKeyHashAddresser::new(prefix, None)
+                    .map_err(|e| format!("Unable to construct Addresser: {}", e))?,
+                _ => DoubleKeyHashAddresser::new(prefix, Some(lengths[0]))
+                    .map_err(|e| format!("Unable to construct Addresser: {}", e))?,
+            };
+            let addr = addresser
+                .compute(&(keys[0].to_string(), keys[1].to_string()))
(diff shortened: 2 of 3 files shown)


examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    // The Addresser implementation to be constructed depends on the amount of `keys` supplied.
+    // Once the appropriate Addresser implementation is constructed, the radix address will be
+    // calculated and printed out.
+    match keys.len() {
+        1 => {
+            let addresser = KeyHashAddresser::new(prefix);
+            let addr = addresser
+                .compute(&keys[0].to_string())
+                .expect("Unable to compute address with KeyHashAddresser");
```
peterschwarz: Change main to 

```
fn main() -> Result<(), String> {
...
```
and `map_err` instead of expect for these.

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+                .map_err(|e| {
+                    format!(
+                        "Unable to compute address with TripleKeyHashAddresser: {}",
+                        e
+                    )
+                })?;
+            println!("{}", addr);
+        }
+        _ => {
+            unreachable!();
```
peterschwarz: Put a comment here about why this is valid. (i.e. it is already checked by clap)

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    match keys.len() {
+        1 => {
+            let addresser = KeyHashAddresser::new(prefix);
+            let addr = addresser
+                .compute(&keys[0].to_string())
+                .map_err(|e| format!("Unable to compute address with KeyHashAddresser: {}", e))?;
+            println!("{}", addr);
+        }
+        2 => {
+            let addresser = DoubleKeyHashAddresser::new(prefix, None);
```
vaporos: Instead of None here, we should accept a command line argument to be able to set it.

examples/address_generator/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "address-generator"
+version = "0.1.0"
+authors = ["Cargill Incorporated"]
+edition = "2018"
+
+[dependencies]
+clap = "2.32"
```
vaporos: This can be simplified to clap = "2"

examples/address_generator/src/main.rs
shannynalayna: I addressed this by adding a `lengths` arg

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+                .required(true)
+                .takes_value(true)
+                .max_values(3)
+                .help(
+                    "the natural keys used to compute the radix address, \
+                     must supply 1 - 3 separate keys",
+                ),
+        )
+        .arg(
+            Arg::with_name("lengths")
```
vaporos: This should be singular 'length'.

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+                .long("lengths")
+                .required(false)
+                .takes_value(true)
+                .max_values(2)
+                .help(
+                    "optional lengths when multiple keys are used to compute the radix address \
```
vaporos: This should start with a capital letter. Try and make it work such that when you have a terminal of with 80 it looks ok.

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    let keys: Vec<_> = matches
+        .values_of("keys")
+        .expect("Unable to get `keys` arg")
+        .collect();
+
+    // Optional lengths used in the construction of the radix address
+    let lengths: Vec<usize> = matches
+        .values_of("lengths")
+        .expect("Unable to get `lengths` arg")
```
vaporos: This is a bug -- currently, if you don't specify --lengths, the example exits with this error.

shannynalayna: Addressed

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+                .multiple(true)
+                .takes_value(true)
+                .max_values(3)
+                .index(2)
+                .help(
+                    "Natural key used to compute the radix address.\nMust supply at least 1 key.",
+                ),
+        )
+        .arg(
+            Arg::with_name("length")
```
vaporos: Should make this one a long option (non-positional).

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+fn main() -> Result<(), String> {
+    let matches = App::new(clap::crate_name!())
+        .version(clap::crate_version!())
+        .author(clap::crate_authors!())
+        .about(clap::crate_description!())
+        .arg(
+            Arg::with_name("prefix")
+                .required(true)
+                .takes_value(true)
+                .help("Prefix of the radix address.")
```
vaporos: Remove the period (following the built-in --help and --version options).

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        )
+        .arg(
+            Arg::with_name("key")
+                .required(true)
+                .multiple(true)
+                .takes_value(true)
+                .max_values(3)
+                .index(2)
+                .help(
+                    "Natural key used to compute the radix address.\nMust supply at least 1 key.",
```
vaporos: Remove the newline and second sentence.

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+                ),
+        )
+        .arg(
+            Arg::with_name("length")
+                .required(false)
+                .max_values(2)
+                .last(true)
+                .index(3)
+                .help(
+                    "Optional length of hashes used to construct the radix address.\nMust be u32.",
```
vaporos: Remove newline and 2nd sentence; u32 holds no meaning to sysadmins, and it is already obvious it will be a number. Remove "Optional" from the description, this will be clear when it is a non-positional flag.

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        .expect("Unable to get `prefix` arg")
+        .to_string();
+
+    let keys: Vec<_> = matches
+        .values_of("key")
+        .expect("Unable to get `key` arg")
+        .collect();
+
+    // Optional lengths used in the construction of the radix address
+    let lengths = match matches.values_of("length") {
```
vaporos: In addition to this, I suggest we also let the user supply the lengths as comma separated, so the following would all be the same:

-L 1 -L 2 -L 3
-L 1,2,3
-L 1,2 -L 3

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+                .help(
+                    "Optional length of hashes used to construct the radix address.\nMust be u32.",
+                ),
+        )
+        .get_matches();
+
+    // The `prefix` will be used to construct the Addresser
+    let prefix = matches
+        .value_of("prefix")
+        .expect("Unable to get `prefix` arg")
```
vaporos: I'd use ' instead of `, because `` in bash causes process execution. (I don't have a concrete example of how exactly the output would have to be mishandled to cause an issue though.)

vaporos: Actually, maybe: "Unable to get argument: prefix"

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        .values_of("key")
+        .expect("Unable to get `key` arg")
+        .collect();
+
+    // Optional lengths used in the construction of the radix address
+    let lengths = match matches.values_of("length") {
+        Some(vals) => vals
+            .map(|l| {
+                l.parse()
+                    .map_err(|_| "Unable to parse `length` arg".to_string())
```
vaporos: Maybe: "Argument must be an integer: length" or "--length argument must be an integer"

examples/address_generator/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    // The Addresser implementation to be constructed depends on the amount of `keys` supplied.
+    // Once the appropriate Addresser implementation is constructed, the radix address will be
+    // calculated and printed out.
+    match keys.len() {
+        1 => {
+            let addresser = KeyHashAddresser::new(prefix);
+            let addr = addresser
+                .compute(&keys[0].to_string())
+                .map_err(|e| format!("Unable to compute address with KeyHashAddresser: {}", e))?;
```
vaporos: Remove reference to KeyHashAddresser here. The CLI user isn't familiar with the code.

vaporos: Same below for DoubleKeyHashAddresser and TripleKeyHashAddresser

examples/address_generator/src/main.rs
shannynalayna: Changed this to your first suggestion.

PR #33, opened 2019-12-27
Add SmartContract trait

shannynalayna:
Adds a basic trait `SmartContract` that will implement the `TransactionHandler` trait. This is necessary so that the `KeyValueTransactionContext` can be used within the `apply` method. This feature is added behind the `contract-handler` trait. 

Also makes a few minor adjustments (adding the three Addresser implementations to the `contract-context-key-value` feature in order to compile the tests in the `context::key_value` module which use all three impls)

### DIFF
--- a/libtransact/src/contract/handler.rs
+++ b/libtransact/src/contract/handler.rs
@@ -1,0 +1,39 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::contract::SmartContract;
+use crate::handler::{ApplyError, TransactionContext, TransactionHandler};
+use crate::protocol::transaction::TransactionPair;
+
+impl<T> TransactionHandler for T
+where
+    T: SmartContract + Send,
+{
+    fn family_name(&self) -> &str {
+        self.get_family_name()
+    }
+
+    fn family_versions(&self) -> &[String] {
+        self.get_family_versions()
+    }
+
+    fn apply(
+        &self,
+        transaction: &TransactionPair,
+        context: &mut dyn TransactionContext,
+    ) -> Result<(), ApplyError> {
+        let contract_context = self.make_context(self.get_addresser(), context);
+        self.apply(transaction, contract_context)
+    }
+}
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -48,6 +48,7 @@ experimental = [
     "contract-address-triple-key-hash",
     "contract-context",
     "contract-context-key-value",
+    "contract-handler",
     "key-value-state",
     "redis-db",
 ]
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -60,5 +60,15 @@ contract-address-key-hash = ["contract-address"]
 contract-address-double-key-hash = ["contract-address"]
 contract-address-triple-key-hash = ["contract-address"]
 contract-context = ["contract", "contract-address"]
-contract-context-key-value = ["contract-context", "key-value-state"]
+contract-context-key-value = [
+    "contract-address-double-key-hash",
+    "contract-address-key-hash",
+    "contract-address-triple-key-hash",
+    "contract-context",
+    "key-value-state"
+]
+contract-handler = [
+    "contract-context-key-value",
+    "contract-address"
+]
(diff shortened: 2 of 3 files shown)


libtransact/src/contract/context/handler.rs
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+use std::hash::Hash;
+
+use crate::contract::address::Addresser;
+use crate::contract::context::key_value::KeyValueTransactionContext;
+use crate::handler::{ApplyError, TransactionContext, TransactionHandler};
+use crate::protocol::transaction::TransactionPair;
+
+pub trait SmartContract: Send {
```
vaporos: This trait should be up one level (in contract/mod.rs).

libtransact/src/contract/context/handler.rs
```diff
@@ -0,0 +1,61 @@
...
+    type Key: Eq + Hash;
+    type Addr: Addresser<Self::Key>;
+
+    fn get_family_name(&self) -> &str;
```
vaporos: This trait should not have get_family_name or get_family_verisons (which are used for TP routing only).

libtransact/src/contract/context/handler.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn get_addresser(&self) -> Self::Addr;
+
+    fn apply<'a>(
+        &self,
+        transaction: &TransactionPair,
+        context: KeyValueTransactionContext<'a, Self::Addr, Self::Key>,
+    ) -> Result<(), ApplyError>;
+}
+
+impl<T> TransactionHandler for T
```
vaporos: I think we will instead want a impl of SmartContract on TransactionHandler (near where transaction handler is defined), since SmartContract will be a subset of the TransactionHandler API.

libtransact/src/contract/context/handler.rs
shannynalayna: Ok, those were only present to implement the corresponding TransactionHandler methods.

libtransact/src/handler/mod.rs
```diff
@@ -140,9 +140,9 @@ pub trait TransactionHandler: Send {
     /// transaction processor upon receiving a TpProcessRequest that the
     /// handler understands and will pass in the TpProcessRequest and an
     /// initialized instance of the Context type.
-    fn apply(
+    fn apply<'a>(
```
peterschwarz: This commit isn't necessary and is a breaking API change

shannynalayna: This change was initially required when the `KeyValueTransactionContext` was being created directly in the apply method. I was getting errors that the lifetimes were inconsistent since the txn ctx was expected to have `'a` lifetime but it is not assigned one in this method. However, I changed the context to be generic so this is no longer necessary in this impl.

libtransact/src/contract/handler.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::contract::address::Addresser;
+use crate::handler::{ApplyError, TransactionContext, TransactionHandler};
+use crate::protocol::transaction::TransactionPair;
+
+pub trait SmartContract: Send {
+    type Key: Eq + Hash;
+    type Addr: Addresser<Self::Key>;
+    type Context;
+
+    fn get_family_name(&self) -> &str;
```
vaporos: get_family_name/get_family_versions should be removed from this trait.

libtransact/src/contract/handler.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    fn get_family_versions(&self) -> &[String];
+
+    fn get_addresser(&self) -> Self::Addr;
+
+    fn make_context<'a>(
```
vaporos: How do you envision get_addresser and make_context being used? These aren't things that we really want to force the smart contract author to need to implement.

libtransact/src/contract/handler.rs
```diff
@@ -0,0 +1,65 @@
...
+    ) -> Self::Context;
+
+    fn apply(
+        &self,
+        transaction: &TransactionPair,
+        context: Self::Context,
+    ) -> Result<(), ApplyError>;
+}
+
+impl<T> TransactionHandler for T
```
vaporos: This should be removed from here (this code shouldn't know about TransactionHandler concepts). To mesh these together without changing the TransactionHandler API (yet), we could create a SmartContractEngine trait, impl SmartContract on that trait, and then create a SmartContractEngineTransactionHandler implementation which pulls it together. I think longer-term, we get rid of TransactionHandler in favor of SmartContractEngine, but this approach would buy us some space to figure out how SmartContractEngine should be defined.

libtransact/src/contract/handler.rs
```diff
@@ -0,0 +1,65 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::hash::Hash;
+
+use crate::contract::address::Addresser;
+use crate::handler::{ApplyError, TransactionContext, TransactionHandler};
+use crate::protocol::transaction::TransactionPair;
+
+pub trait SmartContract: Send {
```
vaporos: "SmartContract" could be defined in the contract/mod.rs since it is fundamental.

shannynalayna:
Closing as this change is outdated

PR #38, opened 2020-01-29
Implement Sqlite Database backend

peterschwarz:
This PR adds an experimental Sqlite database implementation.  

Items to be implemented for moving this feature to stable:

[ ] Allow multiple databases to target a single sqlite database file, either through schemas or namespaced tables.
[ ] Determine best default fore Memory-mapped I/O size
[ ] Autovacuuming setting.

Note:

This PR also introduces a breaking API change to the `Database` trait to have the `Database::get` return a `Result`, instead of ignoring errors.

### DIFF
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -13,7 +13,7 @@ description = """\
 repository = "http://github.com/hyperledger/transact"
 
 [package.metadata.docs.rs]
-features = ["sawtooth-compat", "redis-db"]
+features = ["experimental"]
 
 [dependencies]
 hex = "0.3"
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -50,6 +50,7 @@ experimental = [
     "contract-context-key-value",
     "key-value-state",
     "redis-db",
+    "sqlite-db"
 ]
 sawtooth-compat = ["sawtooth-sdk"]
 ursa-compat = ["ursa"]
@@ -62,4 +63,5 @@ contract-address-triple-key-hash = ["contract-address"]
 contract-context = ["contract", "contract-address"]
 contract-context-key-value = ["contract-context", "key-value-state"]
 key-value-state = []
+sqlite-db = []
 state-merkle = ["cbor-codec"]
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -28,7 +28,10 @@ openssl = "0.10"
 uuid = { version = "0.7", features = ["v4"] }
 sawtooth-sdk = { version = "0.3", optional = true }
 ursa = { version = "0.2.0", optional = true }
+r2d2 = { version = "0.8", optional = true }
+r2d2_sqlite = { version = "0.14", optional = true }
 redis = { version = "0.13.0", default-features = false, optional = true }
+rusqlite = { version = "0.21.0", optional = true }
 
 [dev-dependencies]
 rand_hc = "0.1"
@@ -63,5 +66,5 @@ contract-address-triple-key-hash = ["contract-address"]
 contract-context = ["contract", "contract-address"]
 contract-context-key-value = ["contract-context", "key-value-state"]
 key-value-state = []
-sqlite-db = []
+sqlite-db = ["rusqlite", "r2d2", "r2d2_sqlite"]
 state-merkle = ["cbor-codec"]
--- a/libtransact/src/database/lmdb.rs
+++ b/libtransact/src/database/lmdb.rs
@@ -24,6 +24,7 @@ use std::collections::HashMap;
 use std::path::Path;
 use std::sync::Arc;
 
+use lmdb::error::LmdbResultExt as _;
 use lmdb_zero as lmdb;
 
 use crate::database::error::DatabaseError;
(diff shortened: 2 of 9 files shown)


libtransact/src/database/sqlite.rs
```diff
@@ -0,0 +1,989 @@
...
+//! When using this feature with an in-memory sqlite instance, `DatabaseReader`s cannot be opened
+//! at the same time as an existing `DatabaseWriter`. This is particularly true when using a
+//! database path like `"file::memory:?cache=shared".  In this case, the underlying connection pool
+//! will use the same database instance.
+//!
+//! For a single, unshared memory instances, the database should be created with a connection pool
+//! size of 1, such that creating readers or writers across threads will block until a connection
+//! is available.
+//!
+//! This can be acomplished via
```
vaporos: Typo - acomplished

peterschwarz: Fixed

libtransact/src/database/lmdb.rs
```diff
@@ -145,10 +146,13 @@ pub struct LmdbDatabaseReader<'a> {
...
 impl<'a> DatabaseReader for LmdbDatabaseReader<'a> {
-    fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
+    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, DatabaseError> {
         let access = self.txn.access();
-        let val: Result<&[u8], _> = access.get(&self.db.main, key);
-        val.ok().map(Vec::from)
+        access
+            .get(&self.db.main, key)
+            .to_opt()
+            .map(|res: Option<&[u8]>| res.map(Vec::from))
```
ltseeley: Should this be `opt` instead of `res`? Same for writer struct.

libtransact/Cargo.toml
```diff
@@ -28,7 +28,11 @@ openssl = "0.10"
 uuid = { version = "0.7", features = ["v4"] }
 sawtooth-sdk = { version = "0.3", optional = true }
 ursa = { version = "0.2.0", optional = true }
+r2d2 = { version = "0.8", optional = true }
+r2d2_sqlite = { version = "0.14", optional = true }
 redis = { version = "0.13.0", default-features = false, optional = true }
+rusqlite = { version = "0.21.0", optional = true }
+
```
ltseeley: Is this extra line necessary?

libtransact/src/database/sqlite.rs
```diff
@@ -14,3 +14,778 @@
...
+type SqliteConnection = r2d2::PooledConnection<SqliteConnectionManager>;
+
+#[derive(Clone)]
+pub struct SqliteDatabase {
+    pool: r2d2::Pool<SqliteConnectionManager>,
+}
+
+impl SqliteDatabase {
+    pub fn new(path: &str, indexes: &[&str]) -> Result<Self, DatabaseError> {
+        eprintln!("new database at {}", path);
```
ltseeley: Is `eprintln!` used intentionally?

libtransact/src/database/sqlite.rs
```diff
@@ -14,3 +14,778 @@
...
+
+    /// Asserts that KEY is not in DB's INDEX.
+    fn assert_not_in_index(index: &str, key: u8, db: &dyn Database) {
+        let reader = db.get_reader().expect("unable to get a database reader");
+
+        assert!(reader
+            .index_get(index, &[key])
+            .expect("unable to get value by index")
+            .is_none());
+    }
```
ltseeley: Could these be put in the `database` module for reuse by other DB implementations?

libtransact/src/database/sqlite.rs
```diff
@@ -83,6 +83,21 @@ impl SqliteDatabase {
 
         Ok(Self { pool })
     }
+
+    pub fn vacuum(&self) -> Result<(), SqliteDatabaseError> {
```
ltseeley: The documentation for this method should probably be a bit more detailed since its meaning is specific to Sqlite.

libtransact/src/database/sqlite.rs
```diff
@@ -28,42 +30,106 @@ use super::{
...
+
+    /// Set the connection pool size.
+    pub fn with_connection_pool_size(mut self, pool_size: NonZeroU32) -> Self {
+        self.pool_size = Some(pool_size.get());
+        self
+    }
+
+    /// Set the size used for Memory-Mapped I/O.
+    ///
+    /// This can be disabled by setting the value to `0`
```
ltseeley: Could this instead by disabled using an option (i.e. `None` is disabled)?

peterschwarz: `0` is the sqlite configuration value, which I don't think we necessarily need to hide. If it was an option, then it would need to be `Option<NonZeroU64>`, which feels a little heavyweight for a builder. If it was `Option<u64>`, it would also be disableable via `Some(0)`.

libtransact/src/database/lmdb.rs
peterschwarz: Fixed

libtransact/Cargo.toml
peterschwarz: Removed

libtransact/src/database/sqlite.rs
peterschwarz: Removed (though it was removed in a later commit).

libtransact/src/database/sqlite.rs
peterschwarz: They are not particularly useful functions - I may actual considered removing them altogether.

libtransact/src/database/sqlite.rs
peterschwarz: updated.

PR #39, opened 2020-01-30
Add "state-merkle" rust feature

peterschwarz:
Add a rust feature for the Merkle State implemenation: `"state-merkle"`.  This feature is one by default, as the structs or locations have not changed, and were on by default previously.

Mainly this is a PR that reorganizes the merkle code such that it can be guarded under the above feature.


libtransact/src/state/merkle/change_log.rs
```diff
@@ -19,7 +19,7 @@ use crate::protos::merkle;
 use crate::protos::{FromNative, FromProto, IntoNative, IntoProto, ProtoConversionError};
 use protobuf::Message;
 
-use super::merkle::StateDatabaseError;
+use super::error::StateDatabaseError;
```
vaporos: This was probably intended for the previous commit.

peterschwarz: Not, really - moving this file, the super of this is now `state::merkle`, where as prior to this commit, it was `state` and this module was a peer of `merkle`

PR #43, opened 2020-02-06
Add native .scar representation and loading

ltseeley:
Introduces the `SmartContractArchive` struct, which is the native
representation of a .scar file. This struct can be loaded directly from
a .scar file on the filesystem.

### DIFF
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -48,6 +48,7 @@ experimental = [
     "contract-address-double-key-hash",
     "contract-address-key-hash",
     "contract-address-triple-key-hash",
+    "contract-archive",
     "contract-context",
     "contract-context-key-value",
     "key-value-state",
@@ -60,6 +61,7 @@ contract-address = ["contract"]
 contract-address-key-hash = ["contract-address"]
 contract-address-double-key-hash = ["contract-address"]
 contract-address-triple-key-hash = ["contract-address"]
+contract-archive = ["contract"]
 contract-context = ["contract", "contract-address"]
 contract-context-key-value = ["contract-context", "key-value-state"]
 key-value-state = []
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -13,7 +13,9 @@ description = """\
 repository = "http://github.com/hyperledger/transact"
 
 [dependencies]
+bzip2 = { version = "0.3", optional = true }
 cbor-codec = { version = "0.7", optional = true }
+glob = { version = "0.3", optional = true }
 hex = "0.3"
 libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
@@ -26,13 +28,20 @@ rand = "0.6"
 redis = { version = "0.13.0", default-features = false, optional = true }
 rusqlite = { version = "0.21.0", optional = true }
 sawtooth-sdk = { version = "0.3", optional = true }
+semver = { version = "0.9", optional = true }
+serde = { version = "1.0", optional = true }
+serde_derive = { version = "1.0", optional = true }
+serde_yaml = { version = "0.8", optional = true }
 sha2 = "0.8"
+tar = { version = "0.4", optional = true }
 ursa = { version = "0.2.0", optional = true }
 uuid = { version = "0.7", features = ["v4"] }
 
 [dev-dependencies]
 rand_hc = "0.1"
 sawtooth-xo = "0.3"
+serial_test = "0.3"
+tempdir = "0.3"
 
 [build-dependencies]
 protoc-rust = "2"
@@ -61,7 +70,7 @@ contract-address = ["contract"]
 contract-address-key-hash = ["contract-address"]
 contract-address-double-key-hash = ["contract-address"]
 contract-address-triple-key-hash = ["contract-address"]
-contract-archive = ["contract"]
+contract-archive = ["bzip2", "contract", "glob", "semver", "serde", "serde_derive", "serde_yaml", "tar"]
 contract-context = ["contract", "contract-address"]
 contract-context-key-value = ["contract-context", "key-value-state"]
 key-value-state = []
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -12,26 +12,23 @@ description = """\
 """
 repository = "http://github.com/hyperledger/transact"
 
-[package.metadata.docs.rs]
-features = ["experimental"]
-
 [dependencies]
+cbor-codec = { version = "0.7", optional = true }
 hex = "0.3"
-protobuf = "2"
-rand = "0.6"
-sha2 = "0.8"
+libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
 log = { version = "0.4", features = ["std"] }
-cbor-codec = { version = "0.7", optional = true }
-libc = ">=0.2.35"
 openssl = "0.10"
-uuid = { version = "0.7", features = ["v4"] }
-sawtooth-sdk = { version = "0.3", optional = true }
-ursa = { version = "0.2.0", optional = true }
+protobuf = "2"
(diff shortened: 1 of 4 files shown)


libtransact/Cargo.toml
```diff
@@ -66,5 +63,17 @@ contract-address-triple-key-hash = ["contract-address"]
...
 state-merkle = ["cbor-codec"]
+ursa-compat = ["ursa"]
+
+[package.metadata.docs.rs]
+features = [
+  "stable",
+  "experimental",
+  "sawtooth-compat",
+  "state-merkle",
+  "ursa-compat",
```
agunde406: I think this is going to break the docs on publish.

ltseeley: Removed ursa from the list

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,553 @@
...
+use bzip2::read::BzDecoder;
+use glob::glob;
+use semver::{Version, VersionReq};
+use tar::Archive;
+
+pub use error::Error;
+
+/// Load default location(s) for .scar files.
+pub fn default_scar_paths() -> Vec<PathBuf> {
+    match var_os("SCAR_PATH") {
```
agunde406: Can this environment variable be defined above as a const like `const SCAR_PATH_ENV: &str = "SCAR_PATH";`

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,553 @@
...
+        Some(paths) => split_paths(&paths).collect(),
+        None => vec![PathBuf::from("/usr/share/scar")],
```
agunde406: Can you also define a const for the default location like `const DEFAULT_SCAR_PATH: &str = "/usr/share/scar";

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,553 @@
...
+    /// ```
+    pub fn from_scar_file<P: AsRef<Path>>(
+        name: &str,
+        version: &str,
+        path: P,
+    ) -> Result<SmartContractArchive, Error> {
+        let file_path = find_scar(name, version, path)?;
+
+        let scar_file = File::open(file_path)
+            .map_err(|err| Error::new_with_source("failed to open file", err.into()))?;
```
agunde406: Does this print what file failed to open? As in the file name, not just that the file does not exist.

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,553 @@
...
+        let mut archive = Archive::new(BzDecoder::new(scar_file));
+        let archive_entries = archive
+            .entries()
+            .map_err(|err| Error::new_with_source("failed to read scar file", err.into()))?;
```
agunde406: Same as above this should include what file failed to be read.

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,553 @@
...
+
+        let mut metadata = None;
+        let mut contract = None;
+
+        for entry in archive_entries {
+            let mut entry = entry.map_err(|err| {
+                Error::new_with_source(
+                    "invalid .scar file: failed to read archive entry",
```
agunde406: Please make sure all errors provide enough information that we now which .scar file is failing.

libtransact/src/contract/archive/mod.rs
ltseeley: Updated

libtransact/src/contract/archive/mod.rs
ltseeley: Updated

libtransact/src/contract/archive/mod.rs
ltseeley: Updated

libtransact/src/contract/archive/mod.rs
ltseeley: Updated

libtransact/src/contract/archive/mod.rs
ltseeley: Updated

libtransact/Cargo.toml
```diff
@@ -66,5 +63,16 @@ contract-address-triple-key-hash = ["contract-address"]
...
 key-value-state = []
+redis-db = ["redis"]
+sawtooth-compat = ["sawtooth-sdk"]
 sqlite-db = ["rusqlite", "r2d2", "r2d2_sqlite"]
 state-merkle = ["cbor-codec"]
+ursa-compat = ["ursa"]
+
+[package.metadata.docs.rs]
+features = [
+  "stable",
```
agunde406: This file doesn't have stable.

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,586 @@
...
+            scar_path
+        );
+    }
+
+    /// Verify that the .scar file with no pre-release tag is returned rather than the same patch
+    /// version with a pre-release tag.
+    #[test]
+    fn find_scar_highest_matching_patch() {
+        let dir = new_temp_dir();
+        write_mock_scar(&dir, "mock", "0.1.2-dev");
```
agunde406: Can you add some tests where the -dev version would be picked?

libtransact/Cargo.toml
ltseeley: Fixed (added default and nightly)

libtransact/src/contract/archive/mod.rs
ltseeley: Added (`find_scar_pre_release`)

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,599 @@
...
+        assert_eq!(
+            find_scar("mock", "0.1.2", &dir).expect("failed to find .scar"),
+            scar_path
+        );
+    }
+
+    /// Verify that a pre-release .scar file is correctly found when it meets the version
+    /// requirements.
+    #[test]
+    fn find_scar_pre_release() {
```
agunde406: Looks like this test failed

ltseeley: Removed, updated exact version test to get a pre-release version.

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,586 @@
...
+use semver::{Version, VersionReq};
+use tar::Archive;
+
+pub use error::Error;
+
+const DEFAULT_SCAR_PATH: &str = "/usr/share/scar";
+const SCAR_PATH_ENV_VAR: &str = "SCAR_PATH";
+
+/// Load default location(s) for .scar files.
+pub fn default_scar_paths() -> Vec<PathBuf> {
```
vaporos: This would more naturally be default_scar_path(); common env variables (PATH, LD_RUN_PATH, etc.) are always singular but also always a list.

libtransact/src/contract/archive/mod.rs
```diff
@@ -14,3 +14,586 @@
...
+    /// To load v1.2.3 explicitly, specify the exact version:
+    ///
+    /// ```no_run
+    /// use transact::contract::archive::SmartContractArchive;
+    ///
+    /// let archive = SmartContractArchive::from_scar_file("xo", "1.2.3", "/usr/share/scar")
+    ///     .expect("failed to load .scar file");
+    /// assert_eq!(&archive.metadata.version, "1.2.3");
+    /// ```
+    pub fn from_scar_file<P: AsRef<Path>>(
```
vaporos: load_scar_file() would also make sense here, since it is loading it from the filesystem.

libtransact/src/contract/archive/mod.rs
ltseeley: Updated

PR #45, opened 2020-02-11
Add a configurable prefix for sqlite tables

peterschwarz:
Add a configurable prefix for all of the tables created/used by the SqliteDatabase implementation.  This allows for multiple transact Database instances to be applied to a single sqlite database file.


jsmitchell:
You should consider moving all these queries to prepared statements in a different PR.

jsmitchell:
> You should consider moving all these queries to prepared statements in a different PR.

Or maybe prepare_cached() is already doing that?

peterschwarz:
`prepare_cached` should be doing that. If the format calls are expensive (amortized) we could consider pre-computing the sql strings.

peterschwarz:
Changed the default prefix to `"transact_"` and removed the double prefix.

PR #49, opened 2020-02-19
Add receipt/result conversions to state changes 

ltseeley:
Adds `TryFrom` implementations for both `TransactionResult` and
`TransactionReceipt` to convert them to vectors of state changes.

### DIFF
--- a/libtransact/src/protocol/receipt.rs
+++ b/libtransact/src/protocol/receipt.rs
@@ -14,17 +14,19 @@
  * limitations under the License.
  * -----------------------------------------------------------------------------
  */
+
 //! The `receipts` module contains structs that supply information on the processing
 //! of `Transaction`s
-use protobuf::Message;
 
-use crate::protos;
-use crate::protos::{
-    FromBytes, FromNative, FromProto, IntoBytes, IntoNative, IntoProto, ProtoConversionError,
-};
-use crate::state;
 use std::error::Error as StdError;
 use std::fmt;
+
+use protobuf::Message;
+
+use crate::protos::{
+    self, FromBytes, FromNative, FromProto, IntoBytes, IntoNative, IntoProto, ProtoConversionError,
+};
+use crate::state;
 
 /// A change to be applied to state, in terms of keys and values.
 ///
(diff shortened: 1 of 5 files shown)


libtransact/src/protocol/receipt.rs
```diff
@@ -61,6 +62,61 @@ impl From<StateChange> for state::StateChange {
...
+        let state_changes: Vec<StateChange> = result.try_into()?;
+        Ok(state_changes
+            .into_iter()
+            .map(state::StateChange::from)
+            .collect())
+    }
+}
+
+impl TryFrom<TransactionReceipt> for Vec<StateChange> {
+    type Error = String;
```
vaporos: We should maybe create an error type for this, "InvalidConversionError" which takes the transaction id in its constructor and provides a transaction_id() method. this could then be used by the caller to construct a better error message.

If so, we should do the same for TransactionResult but maybe use an error without the id.

In both cases, because we want to do some processing related to the error, it is probably best to name the errors explicitly so if we change this in the future, we don't introduce errors which could result for another reason without the ability to detect that change occurred.

libtransact/src/protocol/receipt.rs
```diff
@@ -61,6 +62,61 @@ impl From<StateChange> for state::StateChange {
...
+
+    fn try_from(receipt: TransactionReceipt) -> Result<Self, Self::Error> {
+        let TransactionReceipt {
+            transaction_result,
+            transaction_id,
+        } = receipt;
+
+        transaction_result.try_into().map_err(|err| {
+            format!(
+                "failed to convert transaction receipt {}: {}",
```
vaporos: If we introduce an explicit error as suggested above, this code should not use the string conversion on the causing error and instead provide the full error string here. This will cut down one level of error messaging (right now we are getting fairly deep on our error strings for this chaining, so an opportunity like this to cut out a level seems good). (Really, this string construction would be done via the error struct, so this probably amounts to not chaining the errors together at all (source is irrelevant).

libtransact/src/protocol/receipt.rs
ltseeley: Added new error types

libtransact/src/protocol/receipt.rs
ltseeley: The new error types do not chain

PR #53, opened 2020-04-22
Refer to scar files without a . in doc comments and error msgs

agunde406:
There was a mix of "scar file" and ".scar file". This
commit changes the error messages and doc comments to use
"scar file" everywhere.

Signed-off-by: Andrea Gunderson <agunde@bitwise.io>


vaporos:
I think maybe the reverse would be better, instead of ".scar file" we consistently use the non-extension form "scar file".

PR #58, opened 2020-04-30
Enforce .scar file naming restrictions

shannynalayna:
Enforces .scar file or smart contract naming restrictions, which
means the smart contract must not contain a '_' and the version
must be a valid semver string.

An error is returned if these restrictions are not met.

Signed-off-by: Shannyn Telander <telander@bitwise.io>

### DIFF
--- a/libtransact/src/contract/archive/mod.rs
+++ b/libtransact/src/contract/archive/mod.rs
@@ -74,6 +74,8 @@ impl SmartContractArchive {
     /// The scar file to load will be named `<name>_<version>.scar`, where `<name>` is the name
     /// specified for this method and `<version>` is a valid semver string.
     ///
+    /// The `name` argument for this method must not contain any underscores, '_'.
+    ///
     /// The `version` argument for this method must be a valid semver requirement string. The
     /// latest contract (based on semver requirement rules) that matches the given name and meets
     /// the version requirements will be loaded.
@@ -213,6 +215,9 @@ pub struct SmartContractMetadata {
 
 fn find_scar<P: AsRef<Path>>(name: &str, version: &str, paths: &[P]) -> Result<PathBuf, Error> {
     let file_name_pattern = format!("{}_*.scar", name);
+
+    validate_contract_name(name)?;
+
     let version_req = VersionReq::parse(version)?;
 
     // Start with all scar files that match the name, from all paths
@@ -254,6 +259,16 @@ fn find_scar<P: AsRef<Path>>(name: &str, version: &str, paths: &[P]) -> Result<P
                 name, version_req, paths,
             ))
         })
+}
+
+fn validate_contract_name(name: &str) -> Result<(), Error> {
+    if name.contains('_') {
+        return Err(Error::new(&format!(
+            "invalid scar file name, must not include '_': {}",
+            name
+        )));
+    }
+    Ok(())
 }
 
 fn path_is_manifest(path: &std::path::Path) -> bool {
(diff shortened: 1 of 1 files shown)


libtransact/src/contract/archive/mod.rs
```diff
@@ -256,6 +261,21 @@ fn find_scar<P: AsRef<Path>>(name: &str, version: &str, paths: &[P]) -> Result<P
...
+        return Err(Error::new(&format!(
+            "invalid .scar file name, must not include '_': {}",
+            file_name
+        )));
+    } else if let Err(err) = VersionReq::parse(version) {
+        return Err(Error::new_with_source(
+            &format!("invalid version requirement: `{}`", version),
+            Box::new(err),
+        ));
+    }
```
ltseeley, line 275: This is already done in the `find_scar` method, so it's not necessary here.

libtransact/src/contract/archive/mod.rs
```diff
@@ -256,6 +261,21 @@ fn find_scar<P: AsRef<Path>>(name: &str, version: &str, paths: &[P]) -> Result<P
         })
 }
 
+fn validate_scar_file_name(file_name: &str, version: &str) -> Result<(), Error> {
+    if file_name.contains('_') {
+        return Err(Error::new(&format!(
+            "invalid .scar file name, must not include '_': {}",
```
ltseeley, line 267: s/.scar/scar/

libtransact/src/contract/archive/mod.rs
```diff
@@ -325,6 +345,25 @@ mod tests {
...
 
+    /// Verify that an error is returned when the contract name contains '_'.
+    #[test]
+    fn find_scar_invalid_name() {
+        let dir = new_temp_dir();
+
+        assert!(find_scar("non_existent", "0.1.0", &[&dir]).is_err());
+    }
+
+    /// Verify that an error is returned when the contract name contains '_'.
```
ltseeley, line 356: Copypasta

libtransact/src/contract/archive/mod.rs
```diff
@@ -325,6 +345,25 @@ mod tests {
         assert!(find_scar("nonexistent", "0.1.0", &[&dir]).is_err());
     }
 
+    /// Verify that an error is returned when the contract name contains '_'.
```
ltseeley, line 348: For these two tests, you should write a matching scar file with `write_mock_scar` to ensure that the error is being returned because the name/version is invalid, not because the file doesn't exist.

libtransact/src/contract/archive/mod.rs
```diff
@@ -325,6 +345,25 @@ mod tests {
...
+
+    /// Verify that an error is returned when the contract name contains '_'.
+    #[test]
+    fn find_scar_invalid_semver() {
+        let dir = new_temp_dir();
+
+        let err = find_scar("nonexistent", "0..1..0", &[&dir]);
+
+        assert!(err.is_err());
+        assert!(find_scar("nonexistent", "0..1..0", &[&dir]).is_err());
```
ltseeley, line 364: Looks like the same thing is being tested twice.

shannynalayna: Whoops, yeah, forgot to update that test

libtransact/src/contract/archive/mod.rs
shannynalayna: Updated

libtransact/src/contract/archive/mod.rs
shannynalayna: Updated

libtransact/src/contract/archive/mod.rs
shannynalayna: Addressed

PR #66, opened 2020-07-31
Update Sawtooth SDK version 0.3 -> 0.4

shannynalayna:
Also updates the XO example to pull in transact using a relative path rather than w/ a version number and updates the XO example for Sawtooth SDK 0.4.

Also updates one of the sawtooth unit tests to use `HashSigner` instead of just `Signer`.

### DIFF
--- a/examples/simple_xo/Cargo.toml
+++ b/examples/simple_xo/Cargo.toml
@@ -20,8 +20,8 @@ authors = ["IBM Corp.", "Cargill Incorporated"]
 edition = "2018"
 
 [dependencies]
-transact = { version = "0.1.6", features = ["sawtooth-compat"] }
-sawtooth-xo = "0.3"
+transact = { path = "../../libtransact", features = ["sawtooth-compat"] }
+sawtooth-xo = "0.4"
 hex = "0.3"
 sha2 = "0.8"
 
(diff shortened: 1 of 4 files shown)


examples/simple_xo/Cargo.toml
```diff
@@ -20,8 +20,8 @@ authors = ["IBM Corp.", "Cargill Incorporated"]
 edition = "2018"
 
 [dependencies]
-transact = { version = "0.1.6", features = ["sawtooth-compat"] }
-sawtooth-xo = "0.3"
+transact = { path = "../../libtransact", features = ["sawtooth-compat"] }
+sawtooth-xo = "0.4.4"
```
vaporos, line 24: Use 0.4 here instead of 0.4.4 unless there is a reason.

libtransact/src/sawtooth.rs
```diff
@@ -406,7 +406,7 @@ mod xo_compat_test {
         }
     }
 
-    fn create_batch(signer: &Signer, game_name: &str, payload: &str) -> BatchPair {
+    fn create_batch(signer: &HashSigner, game_name: &str, payload: &str) -> BatchPair {
```
vaporos, line 409: If possible, we should keep Signer here for example purposes (so if someone is looking at this, its not tied to HashSigner).

PR #67, opened 2020-08-10
Minor API updates

ltseeley:
- Updates transaction dependencies to be strings instead of bytes for consistency with header signatures

### DIFF
--- a/libtransact/src/execution/adapter/test_adapter.rs
+++ b/libtransact/src/execution/adapter/test_adapter.rs
@@ -233,7 +233,7 @@ mod tests {
     fn make_transaction(signer: &dyn Signer) -> TransactionPair {
         TransactionBuilder::new()
             .with_batcher_public_key(vec![])
-            .with_dependencies(vec![vec![]])
+            .with_dependencies(vec!["".to_string()])
             .with_family_name("test".to_string())
             .with_family_version(FAMILY_VERSION.to_string())
             .with_inputs(vec![vec![]])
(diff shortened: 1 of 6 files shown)


libtransact/src/protocol/batch.rs
```diff
@@ -142,6 +142,10 @@ impl Batch {
         &self.header
     }
 
+    pub fn parse_header(&self) -> Result<BatchHeader, BatchBuildError> {
```
agunde406, line 145: Should the commit say Add `parse_header` ?

ltseeley: Oops, meant to change this. Updated the method to `deserialized_header`.

libtransact/src/protocol/batch.rs
```diff
@@ -142,6 +142,10 @@ impl Batch {
         &self.header
     }
 
+    pub fn deserialized_header(&self) -> Result<BatchHeader, BatchBuildError> {
```
peterschwarz, line 145: I don't think I agree with this change - `BatchPair` and `TransactionPair` is the intended route as it essentially caches the header deserialization.  This function gives the user a route to get the header, but doesn't cache it, so it could create an unexpected frequent expense.

ltseeley: In libsawtooth/sawtooth-core, you're mostly dealing with references to batches and transactions because you're getting them from a block. This means to get the header, you have to clone the whole thing then build it; and that's usually just to get a single field from the header. That's not efficient or ergonomic. It seems there's value in having both options available; we could add a note in a doc string for this method to provide information on when you would use this over `into_pair`.

agunde406 (dismissed):

libtransact/src/protocol/batch.rs
ltseeley: Removed this commit

peterschwarz (approved):

agunde406 (approved):

PR #68, opened 2020-08-13
Add ContextLifecycle::clone_box and `Clone` impl

ltseeley:
Adds the `ContextLifecycle::clone_box` trait method and provides an
implementation of the `Clone` trait for `Box<dyn ContextLifecycle>`.
This allows multiple schedulers to be created by cloning a single
context lifecycle.

### DIFF
--- a/libtransact/src/context/manager/mod.rs
+++ b/libtransact/src/context/manager/mod.rs
@@ -26,6 +26,7 @@ use crate::context::{Context, ContextId, ContextLifecycle};
 use crate::protocol::receipt::{Event, StateChange, TransactionReceipt, TransactionReceiptBuilder};
 use crate::state::Read;
 
+#[derive(Clone)]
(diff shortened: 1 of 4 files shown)


libtransact/src/context/manager/mod.rs
```diff
@@ -60,6 +61,10 @@ impl ContextLifecycle for ContextManager {
             .build()?;
         Ok(new_transaction_receipt)
     }
+
+    fn clone_box(&self) -> Box<dyn ContextLifecycle> {
```
peterschwarz, line 65: This feels a bit wrong with this implementation, as it copies all the content (and therefore doesn't properly operate on any new contexts created in the original object.

peterschwarz: Though, perhaps it creates a bit more expediency on getting #6 correct and merged.

PR #70, opened 2020-08-21
Scheduler interface updates

ltseeley:
Includes the following updates to the scheduler interface:
- Requires implementations of `Scheduler` to be `Send`
- Adds the `SchedulerFactory` trait and implementations for creating generic schedulers
- Updates the `Scheduler::cancel` method to abort/return any in-progress batches and to send a notification if the scheduler is finished
- Updates schedulers to shut themselves down when they are finished

### DIFF
--- a/libtransact/src/scheduler/multi/mod.rs
+++ b/libtransact/src/scheduler/multi/mod.rs
@@ -24,7 +24,8 @@ mod shared;
 
 use crate::protocol::batch::BatchPair;
 use crate::scheduler::{
-    BatchExecutionResult, ExecutionTask, ExecutionTaskCompletionNotifier, Scheduler, SchedulerError,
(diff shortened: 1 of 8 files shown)


libtransact/src/scheduler/multi/mod.rs
```diff
@@ -73,7 +74,7 @@ pub struct MultiScheduler {
 impl MultiScheduler {
     /// Returns a newly created `MultiScheduler` that runs the specified sub-schedulers.
     pub fn new(
-        mut schedulers: Vec<Box<dyn Scheduler + Send>>,
```
peterschwarz, line 76: This looks like it belongs in the Send commit

ltseeley: Good catch; fixed

PR #71, opened 2020-08-24
Scheduler factory fixes

ltseeley:


libtransact/src/scheduler/multi/mod.rs
```diff
@@ -228,16 +229,16 @@ impl MultiSchedulerFactory {
...
             .sub_scheduler_factories
-            .iter_mut()
+            .iter()
             .flat_map(|(factory, num)| {
                 let state_id = state_id.clone();
                 (0..*num).map(move |_| factory.create_scheduler(state_id.clone()))
             })
             .collect::<Result<_, _>>()?;
-        MultiScheduler::new(schedulers, &mut *self.sub_scheduler_handler)
+        MultiScheduler::new(schedulers, &mut **self.sub_scheduler_handler.borrow_mut())
```
peterschwarz, line 241: Won't this byte you if this is called more than once?

ltseeley: Updated to use `try_borrow_mut` to avoid the possibility of a panic.

PR #72, opened 2020-08-26
Replace signing module with cylinder

ltseeley:
Replaces the `transact::signing` module with the `cylinder` library,
which provides a similar interface and implementations.


examples/simple_xo/src/main.rs
```diff
@@ -41,7 +41,7 @@ fn main() {
     let merkle_state = MerkleState::new(db.clone());
     let orig_root = merkle_db.get_merkle_root();
 
-    let signer = HashSigner::new(vec![00u8, 1, 2]);
+    let signer = HashSigner;
```
vaporos, line 44: Now that we can avoid it, we should take HashSigner out of this example - we only had it here out of necessity.

PR #73, opened 2020-08-27
Separate task submission from Executor

peterschwarz:
Prior to this change, the `Executor` could not be shared among threads, requiring the `Executor` to be used by a single thread or guarded by an `Arc<Mutex<_>>`.

This PR separates the task submission from the API of the `Executor`.  This allows for separate threads to submit tasks iterators for processing.  

The `Executor `continues to own the `JoinHandle` for its background thread.

In addition to the above change, this PR:

- Move the error to its own module
- remove the unnecessary use of `AtomicBool` for shutting down various threads.

### DIFF
--- a/libtransact/src/execution/executor/mod.rs
+++ b/libtransact/src/execution/executor/mod.rs
@@ -15,6 +15,7 @@
  * -----------------------------------------------------------------------------
  */
 
+mod error;
 mod internal;
 mod reader;
 
@@ -28,6 +29,8 @@ use crate::scheduler::ExecutionTaskCompletionNotifier;
 use log::debug;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
+
+pub use self::error::ExecutorError;
 
 pub struct Executor {
     readers: Arc<Mutex<HashMap<usize, ExecutionTaskReader>>>,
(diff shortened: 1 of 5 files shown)


libtransact/src/execution/executor/mod.rs
```diff
@@ -100,13 +90,62 @@ impl Executor {
...
-impl SubSchedulerHandler for Executor {
+/// The interface for submitting execution tasks to the Executor.
+#[derive(Clone)]
+pub struct ExecutionTaskSubmitter {
+    sender: Sender<ExecutorCommand>,
+    readers: Arc<Mutex<HashMap<usize, ExecutionTaskReader>>>,
+}
+
+impl ExecutionTaskSubmitter {
+    /// Submit an Iterator of Execution tasks and a completion notifier to the Executor for
```
ltseeley, line 101: s/Submit/Submits/

libtransact/src/execution/executor/mod.rs
```diff
@@ -100,13 +90,62 @@ impl Executor {
...
+    /// processing.  The iterator provided will be consumed until either it is exhausted or the
+    /// Executor is shutdown.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `ExecutorError`if the tasks cannot be successfully sent to the Executor, either
+    /// due to the execution being shutdown .
```
ltseeley, line 108: Missing a space in the first line, and looks like some missing info in the second line.

libtransact/src/execution/executor/mod.rs
peterschwarz: Fixed.

libtransact/src/execution/executor/mod.rs
peterschwarz: Fixed

PR #75, opened 2020-09-01
Fix a bug where None is never returned on SerialScheduler completion

agunde406:
If the scheduler was finalized while there was still unscheduled
batches, None would not be returned. When try_schedule_next
was called, the scheduler would not be ready. Since the task
iterator was empty, next will never be true and try_schedule_next
would never be called again.

This commit checks to see if the all batches have been scheduled
and returns None if finalized before checking if the scheduler
is ready. This will allow the None to always be returned when
complete.

This commit also fixes a bug where if there are no unscheduled
batches the SerialScheduler would always shutdown, even
if not finalized.

Signed-off-by: Andrea Gunderson <agunde@bitwise.io>

### DIFF
--- a/libtransact/src/scheduler/mod.rs
+++ b/libtransact/src/scheduler/mod.rs
@@ -487,6 +487,85 @@ mod tests {
         assert_eq!(result, invalid_receipt_from_batch(batch));
     }
 
+    /// Tests a simple scheduler workflow of processing two batches; this tests the
+    /// scheduler's task iterator, notifier, and result callback functionality.
+    ///
+    /// For the purposes of this test, we simply return an invalid transactions
+    /// as we are not testing the actual execution of the transactions but
+    /// rather the flow of getting a result after adding the batch.
+    ///
+    /// Also verifies that a None is sent to the result callback, once all batches are handled
+    /// if the scheduler is finalized
+    pub fn test_scheduler_flow_with_two_batches(scheduler: &mut dyn Scheduler) {
+        // Use a channel to pass the result to this test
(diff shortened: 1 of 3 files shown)


libtransact/src/scheduler/serial/mod.rs
```diff
@@ -274,6 +274,20 @@ mod tests {
         scheduler.finalize().expect("Failed to finalize");
     }
 
+    /// Tests that a finalized serial scheduler can process two batch with a single transaction and
```
ltseeley, line 277: s/batch/batches/

libtransact/src/scheduler/mod.rs
```diff
@@ -487,6 +487,85 @@ mod tests {
         assert_eq!(result, invalid_receipt_from_batch(batch));
     }
 
+    /// Tests a simple scheduler workflow of processing a two batches; this tests the
```
ltseeley, line 490: s/a //

libtransact/src/scheduler/mod.rs
```diff
@@ -487,6 +487,85 @@ mod tests {
...
+    pub fn test_scheduler_flow_with_two_batches(scheduler: &mut dyn Scheduler) {
+        // Use a channel to pass the result to this test
+        let (tx, rx) = mpsc::channel();
+        scheduler
+            .set_result_callback(Box::new(move |result| {
+                tx.send(result).expect("Failed to send result");
+            }))
+            .expect("Failed to set result callback");
+
+        // Add batch to scheduler
```
ltseeley, line 508: s/batch/batches/

PR #82, opened 2020-09-16
Fix bug with cancelling finalized serial scheduler

ltseeley:
Fixes a bug in the serial scheduler where cancelling the scheduler after
it has been finalized can result in an error. The error occurs when
attempting to receive a possibly aborted batch from the core thread;
since the thread may already have shutdown, the sender could be dropped
immediately. This fixes the error by ignoring the dropped receiver if
the scheduler is already finalized.


peterschwarz (approved):
Ran tests locally

agunde406 (approved):

PR #85, opened 2020-09-21
Manage task readers in executor thread

peterschwarz:
This change manages the task readers in the main internal thread of the Executor.  It allows the executor to fully manage the life-cycle of the task reader threads.

Prior to this change the number of readers would grow over the lifetime of the executor, without being cleaned up.

Additionally, 

- Fix an issue with the task reader indexes
- shutdown all fan-out threads when the loop exits.

### DIFF
--- a/libtransact/src/execution/executor/internal.rs
+++ b/libtransact/src/execution/executor/internal.rs
@@ -301,6 +301,7 @@ impl ExecutorThread {
                 > = HashMap::new();
                 let mut parked: ParkedExecutionEventsMap = HashMap::new();
                 let mut unparked = vec![];
+                let mut reader_index = 0usize;
                 let mut readers: BTreeMap<usize, ExecutionTaskReader> = BTreeMap::new();
                 loop {
                     for execution_event in unparked.drain(0..) {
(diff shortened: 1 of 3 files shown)


vaporos (approved):

vaporos (approved):

libtransact/src/execution/executor/internal.rs
```diff
@@ -344,8 +353,26 @@ impl ExecutorThread {
                                 });
                         }
 
+                        Ok(ExecutorCommand::CreateReader(task_iterator, notifier)) => {
+                            let index = readers.keys().max().cloned().unwrap_or(0);
```
ltseeley, line 357: Shouldn't this be max + 1?

libtransact/src/execution/executor/internal.rs
```diff
@@ -354,6 +381,11 @@ impl ExecutorThread {
                         }
                     }
                 }
+
+                Self::shutdown_fanout_threads(&fanout_threads);
```
ltseeley, line 385: This looks like a separate change; can you explain this in the commit message?

agunde406 (dismissed):

libtransact/src/execution/executor/internal.rs
peterschwarz: It should, though the original code had the same issue - will fix in an additional commit

libtransact/src/execution/executor/internal.rs
peterschwarz: Moved to its own commit.

libtransact/src/execution/executor/internal.rs
agunde406: I had assumed that was on purpose to reset the indexes if they had all been removed.

ltseeley (approved):

agunde406 (approved):

PR #86, opened 2020-09-21
Serial scheduler fix

ltseeley:
Fixes a bug in the serial scheduler that caused the execution task iterator to sometimes hang on shutdown.

### DIFF
--- a/libtransact/src/scheduler/serial/core.rs
+++ b/libtransact/src/scheduler/serial/core.rs
@@ -172,43 +172,57 @@ impl SchedulerCore {
         }
     }
 
+    /// Checks if the scheduler should shutdown, sending the appropriate notifications if necessary.
+    /// The scheduler should shutdown if the following conditions are satisfied:
+    ///
+    /// * The scheduler has been finalized
+    /// * The scheduler is not currently executing a batch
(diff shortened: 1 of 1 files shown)


libtransact/src/scheduler/serial/execution.rs
```diff
@@ -54,11 +54,9 @@ impl Iterator for SerialExecutionTaskIterator {
...
                 }
             }
-            Err(err) => {
-                error!(
-                    "failed to send request for next in execution task iterator: {}",
-                    err
-                );
+            Err(_) => {
+                // This is expected when the scheduler's internal thread has shutdown because it has
+                // no more execution tasks.
```
vaporos, line 59: Instead of ignoring errors (here and above), we should switch to properly sending sentinels which tell us when the other side has shutdown. This way, we still log when this is an error (like the other side crashed or did something unexpected that didn't cleanly shutdown).

ltseeley: I removed this commit and the one that created a new sender for each request; those issues will be addressed in a separate PR.

peterschwarz (approved):

agunde406 (approved):

vaporos (approved):

PR #87, opened 2020-09-23
Improve serial scheduler execution task iterator

ltseeley:
Updates the serial scheduler's execution task iterator to always expect
a `None` result from the internal scheduler thread when the thread
shuts down. If the iterator's send or receive fail and a `None` result
is not received, the thread shutdown incorrectly, so an error is logged
by the iterator. This requires a minor change to the thread's shutdown
procedure: it now sends the `None` result even if the iterator is not
waiting for a new task.

Also logs an error and returns a `None` without requesting a new task if
the iterator has already returned `None`, since no more tasks will be
produced by the scheduler.

### DIFF
--- a/libtransact/src/scheduler/serial/execution.rs
+++ b/libtransact/src/scheduler/serial/execution.rs
@@ -29,11 +29,16 @@ use super::core::CoreMessage;
 pub struct SerialExecutionTaskIterator {
     tx: Sender<CoreMessage>,
     rx: Receiver<Option<ExecutionTask>>,
+    is_complete: bool,
 }
 
 impl SerialExecutionTaskIterator {
     pub fn new(tx: Sender<CoreMessage>, rx: Receiver<Option<ExecutionTask>>) -> Self {
-        SerialExecutionTaskIterator { tx, rx }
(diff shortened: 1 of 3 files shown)


libtransact/src/scheduler/serial/execution.rs
```diff
@@ -42,23 +47,41 @@ impl Iterator for SerialExecutionTaskIterator {
 
     /// Return the next execution task which is available to be executed.
     fn next(&mut self) -> Option<ExecutionTask> {
+        if self.is_complete {
+            error!(
```
peterschwarz, line 51: This should be debug - I don't think this is technically an error condition, as we're not putting the iterator into a state that is invalid (i.e. sending a message for next to the scheduler core when we know there are no more tasks).

vaporos (approved):

libtransact/src/scheduler/serial/execution.rs
ltseeley: Updated to `debug!`

vaporos (approved):

peterschwarz (approved):

PR #89, opened 2020-09-25
Fix serial scheduler task iterator tests

ltseeley:
Updates the `task_iterator_multiple_nones` test to expect any error
instead of a specific error type when verifying that a new task request
was not sent; either `TryReceiveError` variant may be returned depending
on the timing of the test, and both are fine.

Also fixes some lint.

### DIFF
--- a/libtransact/src/scheduler/serial/execution.rs
+++ b/libtransact/src/scheduler/serial/execution.rs
@@ -303,7 +303,7 @@ mod tests {
         fn task_iterator_send_successful_but_receive_failed() {
             let logger = init_logger();
 
-            let (core_tx, core_rx) = channel();
+            let (core_tx, _core_rx) = channel();
             let (_, task_rx) = channel();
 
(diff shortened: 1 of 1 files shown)


agunde406 (approved):

libtransact/src/scheduler/serial/execution.rs
```diff
@@ -303,7 +297,7 @@ mod tests {
         fn task_iterator_send_successful_but_receive_failed() {
             let logger = init_logger();
 
-            let (core_tx, core_rx) = channel();
+            let (core_tx, _core_rx) = channel();
```
vaporos, line 300: Move lint to a separate commit pls.

ltseeley: Updated

agunde406 (approved):

vaporos (approved):

PR #91, opened 2020-10-06
Add state-merkle-redis-db-tests feature

peterschwarz:
This change removes the merkle state tests, backed by Redis from the experimental feature list and puts it behind its own feature flag, leaving only tests that do not rely on external services.  The tests still remain, but must be enabled explicitly.

The `run_tests` script has been updated appropriately, such that the redis-db tests are still run in CI.

### DIFF
--- a/bin/run_tests
+++ b/bin/run_tests
@@ -14,4 +14,5 @@ docker-compose -f docker/compose/docker-compose.yaml run --rm transact \
    /bin/bash -c \
    "cargo test && \
     cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features sawtooth-compat && \
-    cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features experimental"
+    cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features experimental && \
+    cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features experimental,state-merkle-redis-db-tests"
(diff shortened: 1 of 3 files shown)


libtransact/Cargo.toml
```diff
@@ -110,6 +110,7 @@ redis-db = ["redis"]
 sawtooth-compat = ["sawtooth-sdk"]
 sqlite-db = ["rusqlite", "r2d2", "r2d2_sqlite"]
 state-merkle = ["cbor-codec"]
+state-merkle-redis-db-tests = ["redis-db", "state-merkle"]
```
vaporos, line 113: Maybe put a comment on this so its clear this is one one that you need to specify to run full redis tests?

bin/run_tests
```diff
@@ -14,4 +14,5 @@ docker-compose -f docker/compose/docker-compose.yaml run --rm transact \
    /bin/bash -c \
    "cargo test && \
     cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features sawtooth-compat && \
-    cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features experimental"
+    cargo test --manifest-path /project/transact/libtransact/Cargo.toml --features experimental && \
+    cargo test --manifest-path /project/transact/libtransact/Cargo.toml --no-default-features --features state-merkle-redis-db-tests"
```
vaporos, line 18: This should probably be run with experimental features so we know its compatible with other experimental changes (and not run it at all with --no-default-features).

peterschwarz: I did catch a dependency issue by doing this, but I can add back experimental features.

libtransact/Cargo.toml
peterschwarz: added

vaporos: typo - "merlke"

peterschwarz: Fixd

vaporos (approved):

ltseeley (approved):

PR #101, opened 2021-01-18
wasm unknown unknown support

vaporos:
*Updated:*

This PR allows wasm compiles using --no-default-features:

```
cargo build --target wasm32-unknown-unknown --no-default-features
```

It also works when the handler feature is enabled, which would be common for smart contracts:

```
cargo build --target wasm32-unknown-unknown --no-default-features --features=handler
```


agunde406 (approved):

libtransact/Cargo.toml
```diff
@@ -30,12 +30,12 @@ repository = "http://github.com/hyperledger/transact"
...
 bzip2 = { version = "0.3", optional = true }
 cbor-codec = { version = "0.7", optional = true }
-cylinder = "0.2"
+cylinder = { version = "0.2", optional = true }
 glob = { version = "0.3", optional = true }
 hex = "0.3"
 libc = ">=0.2.35"
 lmdb-zero = ">=0.4.1"
-log = { version = "0.4", features = ["std"] }
+log = { version = "0.4", optional = true, features = ["std"] }
```
peterschwarz, line 38: This should also be mentioned in the commit message

peterschwarz (approved):

libtransact/Cargo.toml
vaporos: Fixed

vaporos:
Added a commit which fixes the handler feature.

libtransact/src/protocol/batch.rs
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright 2018 Bitwise IO, Inc.
+ * Copyright 2021 Cargill Incorporated
```
agunde406, line 3: copyright updates should be added to the other commit that updates all of the files touched in this PR

agunde406 (changes requested):

libtransact/src/protocol/batch.rs
agunde406: They are included in the handler fix commit

vaporos: This was intentional because the copyright addition is new instead of being an update.

agunde406 (approved):

agunde406 (approved):

PR #105, opened 2021-01-22
Implement Smallbank Workloads 

agunde406:
This PR starts by merging in the history for the [smallbank_workload](https://github.com/hyperledger/sawtooth-core/tree/master/perf/smallbank_workload) from sawtooth-core.

The workload and playlists are converted from the sawtooth-core implementation to work with transact as well as return Sabre transactions for the smallbank payloads that are generated. These playlist and payloads will be used by future components to submit the workloads to existing networks.

Instead of pulling in the sabre_sdk to create the sabre transaction, a protocol is added for creating ExecuteContractAction Sabre transactions. This was added because the sabre_sdk would have pulled in an old version of transact into master. This is not an issue when using sabre_sdk to build sabre smart contracts because transact is not pulled when compiling into wasm.


libtransact/Cargo.toml
```diff
@@ -53,6 +53,7 @@ serde_yaml = { version = "0.8", optional = true }
 sha2 = "0.8"
 tar = { version = "0.4", optional = true }
 uuid = { version = "0.7", optional = true, features = ["v4"] }
+yaml-rust =  { version = "0.4", optional = true }
```
ltseeley, line 56: Is there a reason for using `yaml-rust` over `serde_yaml`?

agunde406: its what was by the original smallbank_workload code.

ltseeley (approved):

shannynalayna (approved):

PR #106, opened 2021-01-29
Add WorkloadRunner to run BatchWorkloads

agunde406:
The WorkloadRunner makes it easy to run multiple workloads
against different groups of URLS. The workload groups can
be started and removed dynamically, allowing for more
complicated workload test.

This required updating the TransactionWorkload and
BatchWorkload traits to require they can be Send because
the workloads will be run in their own thread.

This is behind an experimental feature "workload-runner"

### DIFF
--- a/libtransact/src/lib.rs
+++ b/libtransact/src/lib.rs
@@ -147,6 +147,6 @@ extern crate log;
 #[cfg(feature = "sabre-compat")]
 #[macro_use]
 extern crate sabre_sdk;
-#[cfg(feature = "contract-archive")]
+#[cfg(feature = "serde_derive")]
 #[macro_use]
 extern crate serde_derive;
--- a/libtransact/src/workload/error.rs
+++ b/libtransact/src/workload/error.rs
@@ -48,3 +48,38 @@ impl From<TransactionBuildError> for WorkloadError {
         WorkloadError::TransactionBuildError(err)
     }
 }
+
+#[cfg(feature = "workload-runner")]
+#[derive(Debug, PartialEq)]
+pub enum WorkloadRunnerError {
+    /// Error raised when failing to submit the batch
+    SubmitError(String),
+    TooManyRequests,
+    /// Error raised when adding workload to the runner
+    WorkloadAddError(String),
+    /// Error raised when removing workload from the runner
+    WorkloadRemoveError(String),
+}
+
+#[cfg(feature = "workload-runner")]
+impl std::error::Error for WorkloadRunnerError {}
+
+#[cfg(feature = "workload-runner")]
+impl std::fmt::Display for WorkloadRunnerError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match *self {
+            WorkloadRunnerError::SubmitError(ref err) => {
+                write!(f, "Unable to submit batch: {}", err)
+            }
+            WorkloadRunnerError::TooManyRequests => {
+                write!(f, "Unable to submit batch because of TooManyRequests")
+            }
+            WorkloadRunnerError::WorkloadAddError(ref err) => {
+                write!(f, "Unable to add workload: {}", err)
+            }
+            WorkloadRunnerError::WorkloadRemoveError(ref err) => {
+                write!(f, "Unable to remove workload: {}", err)
+            }
+        }
+    }
+}
--- a/libtransact/src/workload/error.rs
+++ b/libtransact/src/workload/error.rs
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2018 Bitwise IO, Inc.
+ * Copyright 2019-2021 Cargill Incorporated
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------------
+ */
+
 use crate::protocol::batch::BatchBuildError;
 use crate::protocol::transaction::TransactionBuildError;
 
--- a/libtransact/src/workload/runner.rs
+++ b/libtransact/src/workload/runner.rs
@@ -1,0 +1,423 @@
+/*
+ * Copyright 2021 Cargill Incorporated
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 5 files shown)


libtransact/src/lib.rs
```diff
@@ -147,6 +147,6 @@ extern crate log;
 #[cfg(feature = "sabre-compat")]
 #[macro_use]
 extern crate sabre_sdk;
-#[cfg(feature = "contract-archive")]
+#[cfg(any(feature = "contract-archive", feature = "workload-runner"))]
```
ltseeley, line 150: Can probably just change this to `#[cfg(feature = "serde_derive")]` to avoid adding each feature that depends on that crate.

libtransact/src/workload/error.rs
```diff
@@ -48,3 +48,38 @@ impl From<TransactionBuildError> for WorkloadError {
         WorkloadError::TransactionBuildError(err)
     }
 }
+
+#[cfg(feature = "workload-runner")]
+#[derive(Debug, PartialEq)]
+pub enum WorkloadRunnerError {
+    /// Error raised when fail to submit the batch
```
ltseeley, line 73: s/fail/failing/

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------------
+ */
+
+//! Provides a runner to submit `BatchWorkload`s
+
+// TODO add builder for Worker
```
ltseeley, line 20: TODO

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+use std::{thread, time};
+
+use reqwest::{blocking::Client, header, StatusCode};
+
+use crate::protos::IntoBytes;
+
+use super::error::WorkloadRunnerError;
+use super::BatchWorkload;
+
+const DEFAULT_LOG_TIME: u32 = 5; // time in seconds
```
ltseeley, line 35: Would be helpful to add `_SECS` to the const name so you don't have to see this comment to know what it represents.

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// * `rate`- How many tranactions per second to submit
+    /// * `auth` - The string to be set in the Authorization header for the request
+    ///
+    /// Returns an error if a workload with that ID is already running or if the workload thread
+    /// could not be started
+    pub fn add_workload(
+        &mut self,
+        id: &str,
+        workload: Box<dyn BatchWorkload>,
+        targets: &[String],
```
ltseeley, line 64: Does a workload generate transactions that are dependent on previous transactions? Would that be a problem if a single workload sends transactions to different targets that may not have the dependencies?

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    /// Sets the targets for the worker
+    ///
+    /// # Arguments
+    ///
+    ///  * `targets` - A list of URL for submitting the batches. The URL provided must be the full
+    ///               URL before adding `/batches` for submission
+    pub fn with_targets(mut self, targets: &[String]) -> WorkerBuilder {
+        self.targets = Some(targets.to_vec());
```
ltseeley, line 190: The argument should just be `Vec<String>` to prevent an unnecessary clone. If the calling code doesn't need the vec anymore, it can just pass ownership; it should be up to the caller to determine if a clone is needed.

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+
+        let auth = self.auth.ok_or_else(|| {
+            WorkloadRunnerError::WorkloadAddError(
+                "unable to build, missing field: `auth`".to_string(),
+            )
+        })?;
+
+        let (sender, receiver) = channel();
+
+        let time_to_wait = time::Duration::new(0, 1_000_000_000 / rate as u32);
```
ltseeley, line 247: What is this `1_000_000_000` number? Should it be a `const`?

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+                                let batch = workload.next_batch().expect("Unable to get batch");
+                                let batch_bytes = match vec![batch.batch().clone()].into_bytes() {
+                                    Ok(bytes) => bytes,
+                                    Err(err) => {
+                                        error!("Unable to get batch bytes {}", err);
+                                        break;
+                                    }
+                                };
+
+                                // submit match to the target
```
ltseeley, line 287: s/match/batch/

libtransact/src/workload/runner.rs
peterschwarz: I'm not sure that it's the workload generators concern.  It probably should be up to the targets to ensure that they are exchanging batches in a way such that the network as a whole has all of the dependencies (e.g. like how Sawtooth does it).

libtransact/src/workload/runner.rs
peterschwarz: This is true with the methods that take `&str` as well - a builder is the equivalent of a constructor, so it should really should indicate that it is taking ownership of the values (which it will ultimately pass to the final constructed struct).

libtransact/src/workload/runner.rs
```diff
@@ -0,0 +1,425 @@
...
+
+// TODO add builder for Worker
+
+use std::collections::HashMap;
+use std::fmt;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::mpsc::{channel, Sender, TryRecvError};
+use std::{thread, time};
+
+use reqwest::{blocking::Client, header, StatusCode};
```
peterschwarz, line 28: I wonder if we should consider using the async Client here.

libtransact/src/lib.rs
agunde406: fixed

libtransact/src/workload/error.rs
agunde406: fixed

libtransact/src/workload/runner.rs
agunde406: removed

libtransact/src/workload/runner.rs
agunde406: updated

libtransact/src/workload/runner.rs
agunde406: updated

libtransact/src/workload/runner.rs
agunde406: updated to `let time_to_wait = time::Duration::from_secs(1) / rate;` for clarity

libtransact/src/workload/runner.rs
agunde406: fixed

ltseeley (approved):

peterschwarz (approved):

PR #112, opened 2021-02-12
Make smallbank workload re-runnable 

agunde406:
Before the number of accounts were generated with
client ids in order from 0 to number account requested.
This made it impossible to rerun the workload against
the same network with out having at least the number
of accounts invalid transactions. Instead, the client
id are now randomly generated.

If a seed is provided, the output will be generated the
same way, causing invalid transactions. To be rerun
without the invalid transactions, run again with a
different or no seed to create new accounts.

### DIFF
--- a/cli/src/action/workload.rs
+++ b/cli/src/action/workload.rs
@@ -93,10 +93,13 @@ impl Action for WorkloadAction {
 
         match workload {
             "smallbank" => {
-                let seed = match args.value_of("seed").map(str::parse).unwrap_or_else(|| {
-                    let mut rng = rand::thread_rng();
-                    Ok(rng.gen::<u64>())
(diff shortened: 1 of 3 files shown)


ltseeley (approved):

peterschwarz (commented):
Should the workload fail if an account already exists, or should it just add the account to the list of accounts and continue?

agunde406:
> Should the workload fail if an account already exists, or should it just add the account to the list of accounts and continue?

The workload generator does not wait to find out if the transaction was valid only that it was accepted. It should add the account to the account list if it already exists (the only reason it would be invalid at this time) because the other operations should be valid after account creation is finished.

peterschwarz (approved):

PR #120, opened 2021-03-26
Rust 1.51.0 Clippy Error fixes

agunde406:


vaporos (commented):
Since this breaks APIs, we should consider how/when to apply it. Currently Transact this is 0.3.x. We could create a 0-3 branch, moving main to 0.4 and then apply this to main. In the 0-3 branch we could use allow statements in place of the commits which break APIs.

vaporos (approved):

peterschwarz (approved):

PR #136, opened 2021-05-17
Change wait time between batch submit responsively

Caleb-Hill:
Transact waits the full time between the end of one batch submit and the
next regardless of how long the submit takes. This can make the target
submission rate and the actual submission rate be very different.
This change alters the time between submissions based on how long the
current submission took and a "best guess" for how long they usually
take.
Ideally this will more tightly tie the target and actual submission rate.

Signed-off-by: Caleb Hill <hill@bitwise.io>

### DIFF
--- a/libtransact/src/workload/runner.rs
+++ b/libtransact/src/workload/runner.rs
@@ -276,6 +276,8 @@ impl WorkerBuilder {
                     let start_time = time::Instant::now();
                     // total number of batches that have been submitted
                     let mut submitted_batches = 0;
+                    let mut submission_start = time::Instant::now();
+                    let mut submission_avg: Option<time::Duration> = None;
                     loop {
                         match receiver.try_recv() {
                             // recieved shutdown
@@ -338,7 +340,20 @@ impl WorkerBuilder {
 
                                 // get next target, round robin
                                 next_target = (next_target + 1) % targets.len();
-                                thread::sleep(time_to_wait);
+                                let diff = time::Instant::now() - submission_start;
+                                let submission_time = match submission_avg {
+                                    Some(val) => (diff + val) / 2,
+                                    None => diff,
+                                };
+                                submission_avg = Some(submission_time);
+
+                                let wait_time = match time_to_wait.checked_sub(submission_time) {
+                                    Some(dur) => dur,
+                                    None => time::Duration::from_nanos(0),
+                                };
+
+                                thread::sleep(wait_time);
+                                submission_start = time::Instant::now();
                             }
 
                             Err(TryRecvError::Disconnected) => {
@@ -507,6 +522,8 @@ pub fn submit_batches_from_source(
     let http_counter = HttpRequestCounter::new(format!("File: {}", input_file));
     // the last time http request information was logged
     let mut last_log_time = time::Instant::now();
+    let mut submission_start = time::Instant::now();
+    let mut submission_avg: Option<time::Duration> = None;
     loop {
         let target = match targets.get(next_target) {
             Some(target) => target,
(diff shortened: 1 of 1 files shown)


libtransact/src/workload/runner.rs
```diff
@@ -33,6 +33,7 @@ use super::error::WorkloadRunnerError;
 use super::BatchWorkload;
 
 pub const DEFAULT_LOG_TIME_SECS: u32 = 30; // time in seconds
+const SMOOTHING_FACTOR: f32 = 0.5; //how much to smooth transaction time estimates
```
vaporos, line 36: Add a space between // and "how"

libtransact/src/workload/runner.rs
```diff
@@ -338,7 +341,18 @@ impl WorkerBuilder {
 
                                 // get next target, round robin
                                 next_target = (next_target + 1) % targets.len();
-                                thread::sleep(time_to_wait);
+                                let diff = time::Instant::now() - submission_start;
+                                let submission_time = match submission_avg {
+                                    Some(val) => {
```
vaporos, line 346: This looks like it reduces to avg=(diff+avg)/2; suggest we simplify the implementation here until we need a more complex approach.

vaporos (changes requested):
There is a typo in the commit "bath" instead of "batch" and a empty line missing between paragraphs.

libtransact/src/workload/runner.rs
Caleb-Hill: Simplified to just that expression

libtransact/src/workload/runner.rs
```diff
@@ -338,7 +340,15 @@ impl WorkerBuilder {
...
                                 next_target = (next_target + 1) % targets.len();
-                                thread::sleep(time_to_wait);
+                                let diff = time::Instant::now() - submission_start;
+                                let submission_time = match submission_avg {
+                                    Some(val) => (diff + val) / 2,
+                                    None => diff,
+                                };
+                                submission_avg = Some(submission_time);
+
+                                thread::sleep(time_to_wait - submission_time);
```
agunde406, line 350: I see this throwing the following panic `thread 'Smallbank-Workload-0' panicked at 'overflow when subtracting durations', library/core/src/time.rs:890:31`

agunde406: this happens when the TooManyRequest loop has happened.

agunde406 (changes requested):

libtransact/src/workload/runner.rs
Caleb-Hill: Yup, should have seen that coming. Changed it to a saturating subtraction so the floor is zero always.

libtransact/src/workload/runner.rs
```diff
@@ -554,7 +566,15 @@ pub fn submit_batches_from_source(
...
         next_target = (next_target + 1) % targets.len();
-        thread::sleep(time_to_wait);
+        let diff = time::Instant::now() - submission_start;
+        let submission_time = match submission_avg {
+            Some(val) => (diff + val) / 2,
+            None => diff,
+        };
+        submission_avg = Some(submission_time);
+
+        thread::sleep(time_to_wait - submission_time);
```
agunde406, line 576: This should also be updated to use time_to_wait.saturating_sub

agunde406 (changes requested):

libtransact/src/workload/runner.rs
Caleb-Hill: Yeah, pushed it and then saw the issue. Just pushed yet another fix. Thanks

agunde406 (approved):

vaporos (approved):

PR #137, opened 2021-05-18
Fix Transact docker builds with Docker Compose 1.28+

rbuysse:


docker/compose/.env
```diff
@@ -0,0 +1,3 @@
+ISOLATION_ID=latest
+REPO_VERSION=0.3.7-dev
```
agunde406, line 2: Is this the right version?

agunde406: o i guess it does match the .env file, I though it would be 0.4.1

rbuysse: Something definitely went wrong there. Fixed now.

agunde406 (dismissed):

peterschwarz (approved):

agunde406 (approved):

peterschwarz (approved):

PR #142, opened 2021-05-26
TooManyRequests response bug fix

isabeltomb:
This PR fixes a bug in the recent update to handle `TooManyRequests`. Previously if control-c was pressed while the `slow_rate` method was attempting to resubmit a batch the workload would not stop until `slow_rate` had returned and it would usually result in a panic caused by a divide by zero error.

The following changes are added in this PR to fix this bug:
- After each time `slow_rate` is called the time and total batches counters used to calculate the effective submission rate are reset
- The `slow_rate` method is updated to default to 1 second of sleep if the effective rate is calculated to be 0, this avoids any possible divide by zero errors that were possible before
- A `receiver` argument is added to the `slow_rate` method, the method uses the receiver to check if a shutdown message has been sent each time it loops
- The `slow_rate` method is updated to return a boolean representing whether or not a shutdown message was sent. If `slow_rate` returns `true` the loop will break and the workload will stop

### **Testing:**
1. Start two splinter nodes with the experimental feature "back-pressure"
2. Create a circuit between the nodes, ensure that the scabbard version is set to 2 when creating the circuit
3. Use scabbard CLI to upload the smallbank smart contract
4. Use the transact CLI to start a smallbank workload, for example:
```
transact workload --targets http://localhost:8085/scabbard/<circuit-id>/<service-id> \
--key <private-key-path> \
--target-rate 5 \
--update 2 \
--workload smallbank \
-vv
```
5. Control-c immediately after seeing the log message:
```Received TooManyRequests message from target, attempting to resubmit batch```
check that the log shows:
```
Shutting down worker Smallbank-Workload-0
Worker received shutdown
```
and the workload stops
6. Run the workload command again
7. Control-c while batches are being successfully submitted
check that the same shutdown log message shows:
```
Shutting down worker Smallbank-Workload-0
Worker received shutdown
```
and the workload stops

### DIFF
--- a/libtransact/src/workload/runner.rs
+++ b/libtransact/src/workload/runner.rs
@@ -273,7 +273,7 @@ impl WorkerBuilder {
                     let http_counter = HttpRequestCounter::new(thread_id.to_string());
                     // the last time http request information was logged
                     let mut last_log_time = time::Instant::now();
-                    let start_time = time::Instant::now();
+                    let mut start_time = time::Instant::now();
                     // total number of batches that have been submitted
                     let mut submitted_batches = 0;
                     let mut submission_start = time::Instant::now();
@@ -324,7 +324,8 @@ impl WorkerBuilder {
                                                 submitted_batches,
                                             ) {
                                                 Ok(()) => {
-                                                    submitted_batches += 1;
+                                                    submitted_batches = 1;
+                                                    start_time = time::Instant::now();
                                                     http_counter.increment_sent()
                                                 }
                                                 Err(err) => error!("{}:{}", thread_id, err),
--- a/libtransact/src/workload/runner.rs
+++ b/libtransact/src/workload/runner.rs
@@ -21,7 +21,7 @@ use std::collections::HashMap;
 use std::fmt;
 use std::io::Read;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::mpsc::{channel, Sender, TryRecvError};
+use std::sync::mpsc::{channel, Receiver, Sender, TryRecvError};
 use std::{thread, time};
 
 use reqwest::{blocking::Client, header, StatusCode};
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by agunde406 — DISMISSED** [review]
Tested

**Review by shannynalayna — DISMISSED** [review]

File: libtransact/src/workload/runner.rs [hunk 1]
```diff
@@ -389,9 +389,19 @@ fn slow_rate(
     submitted_batches: u32,
 ) -> Result<(), WorkloadRunnerError> {
     debug!("Received TooManyRequests message from target, attempting to resubmit batch");
-    // figure out the rate at which batches were successfully submitted
-    let effective_rate = submitted_batches / (time::Instant::now() - start_time).as_secs() as u32;
-    let wait = time::Duration::from_secs(1) / effective_rate;
+    let mut shutdown = false;
```
peterschwarz, line 397: This looks like it might be in the wrong commit.

libtransact/src/workload/runner.rs
```diff
@@ -322,12 +322,16 @@ impl WorkerBuilder {
...
                                                 start_time,
                                                 submitted_batches,
+                                                &receiver,
                                             ) {
-                                                Ok(()) => {
-                                                    submitted_batches = 1;
-                                                    start_time = time::Instant::now();
-                                                    http_counter.increment_sent()
-                                                }
+                                                Ok(shutdown) => match shutdown {
```
peterschwarz, line 327: This could be

```rust
Ok(true) => break,
Ok(false) => {
    // false stuff
}
```

peterschwarz (changes requested):

libtransact/src/workload/runner.rs
isabeltomb: Fixed

libtransact/src/workload/runner.rs
isabeltomb: Moved to the correct commit

vaporos (approved):

peterschwarz (approved):

agunde406 (approved):

PR #148, opened 2021-05-28
Remove the clean dependency from just lint

vaporos:
To get the old behavior, run "just clean; just build". This change makes
the clean optional, which is a more desirable developer workflow.

### DIFF
--- a/justfile
+++ b/justfile
@@ -103,7 +103,7 @@ doc:
     set -e
     cargo doc
 
-lint: clean
+lint:
     #!/usr/bin/env sh
     set -e
     echo "\033[1mcargo fmt -- --check\033[0m"


justfile
```diff
@@ -103,7 +103,7 @@ doc:
     set -e
     cargo doc
 
-lint: clean
+lint:
```
agunde406, line 106: This commit removes clean from `just lint` but the commit message says it removes it from `just build`

agunde406 (changes requested):

justfile
vaporos: fixed

peterschwarz (approved):

agunde406 (approved):

PR #178, opened 2021-08-30
Add target rate default from documentation

leebradley:
Signed-off-by: Lee Bradley <bradley@bitwise.io>

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -75,7 +75,6 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
                     Arg::with_name("target_rate")
                         .long("target-rate")
                         .takes_value(true)
-                        .required(true)
                         .long_help(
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -75,7 +75,7 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
                     Arg::with_name("target_rate")
                         .long("target-rate")
                         .takes_value(true)
-                        .required(true)
+                        .default_value("1/s")
```
peterschwarz, line 78: Usually this is done as a `get_value(...).unwrap_or_else("...")`

leebradley: Ah, I see it's already there but it was being blocked by the required(true) clap option. I've updated the PR to just remove this line.

peterschwarz (approved):

vaporos (approved):

PR #220, opened 2021-10-15
Update workload to check batch result

isabeltomb:
- Update `TransactionWorkload`'s `next_transaction` and `BatchWorkload`'s `next_batch` to return an optional `ExpectedBatchResult`. ExpectedBatchResult is an enum with variants `Valid` and Invalid` to indicate the expected outcome of a batch after it has been submitted to a target.
- Update the `next_transaction` and `next_batch` methods in the command implementation of `TransactionWorkload`  and `BatchWorkload` to return an `ExpectedBatchResult` based on the type of transaction being submitted. The `next_transaction` and `next_batch` methods in the smallbank implementation were updated to return None for now as it is not possible to determine the expected result of an amalgamate transaction.
- Add an arg `get_batch_status` to `add_workload` and pass this arg's value to the `WorkerBuilder` `get_batch_status` method. This arg is a boolean value that determines if the worker will check the status of batches after they are submitted to determine if the expected outcome was achieved. This value is set to true when starting a command workload and false when starting a smallbank workload because smallbank does not report the expected outcome of transactions as mentioned above.
- Update the `submit_batch` method to deserialize the REST API response to get the batch status link and return it. The returned link is stored as the key in a HashMap mapped to the target the batch was submitted to and the expected batch result.
- Add a method `check_batch_status` that takes the HashMap containing the batch status links and expected batch results and loops through the links checking if the batch status matches what was expected. If the status link returns the expected result the link is removed from the list so that it will not be checked again. If the status is pending the link will be left in the list so that it can be checked again on the next iteration. `check_batch_status` is called at the end of the loop in the `WorkerBuilder` `build` method if `get_batch_status` is true.

If a command workload is being run and the batch result doesn't match the expected batch result that was set the workload will stop and report an error like the following:
`Unable to submit batch: Expected valid result, received invalid [InvalidTransaction { transaction_id: "5b933e612a09382b2054b7cb53a8bb2f6c1abc3fcb797a8dff08ac6233317fdf279c60d733af0169e3f0bc85a4de5676c07665ce70bfea824702775c0dc8909e", error_message: "Wasm contract returned invalid transaction: command, 1.0", error_data: [] }]`

### DIFF
--- a/libtransact/src/workload/runner.rs
+++ b/libtransact/src/workload/runner.rs
@@ -22,7 +22,7 @@ use std::fmt;
 use std::io::Read;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::mpsc::{channel, Receiver, Sender, TryRecvError};
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 use std::{thread, time};
 
 use reqwest::{blocking::Client, header, StatusCode};
@@ -273,6 +273,8 @@ impl WorkerBuilder {
             )
         })?;
 
+        let get_batch_status = self.get_batch_status.unwrap_or(false);
+
         let update_time = self.update_time.unwrap_or(DEFAULT_LOG_TIME_SECS);
 
         let (sender, receiver) = channel();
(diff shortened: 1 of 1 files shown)


libtransact/src/workload/runner.rs
```diff
@@ -521,12 +532,142 @@ fn submit_batch(
         })
 }
 
+fn check_batch_status(
+    mut status_links: ExpectedBatchResults,
+    auth: &str,
+) -> Result<ExpectedBatchResults, WorkloadRunnerError> {
+    for (status_link, (target, expected_result)) in status_links.clone() {
+        // remove the part of the target URL that is in the beginning of the status link
+        let mut parts = target.splitn(2, "/scabbard");
```
vaporos, line 541: libtransact shouldn't know about scabbard specifically.

isabeltomb: Created a method that will remove overlapping parts of the target/status link instead of specifically using the "/scabbard" substring

PR #249, opened 2021-11-03
`workload-runner` stabilization updates

isabeltomb:
This PR fixes a bug where the break statements in the `WorkerBuilder` `build` function didn't properly shutdown the workload. This bug is fixed by switching from using an AtomicBool to a shutdown signaler. Previously an AtomicBool was used to indicate the running state of a workload and a loop was used to keep the workloads running until a sigint was received. With this setup, any break statements in the `WorkerBuilder` `build` function would exit the inner loop and cause the program to hang because the outer loop would not exit until the AtomicBool was changed by the ctrlc handler. Using the shutdown signaler and the `wait_for_shutdown` function allow for the workload to shutdown properly if any errors are encountered. As a result of this change the duration value is passed to the `add_workload` function and used in the `build` function of `WorkerBuilder` to calculate the end time and stop the workload when the end time is reached.

Other stabilization related changes included in this PR:
- Remove an `expect` in the worker thread
- Replace `exit` with `break` in the worker thread


vaporos (approved):

vaporos (approved):

agunde406 (approved):

cli/src/action/workload.rs
```diff
@@ -170,30 +168,21 @@ impl Action for WorkloadAction {
             }
         }
 
-        // calculate the end time based on the duration given
-        let end_time = duration.map(|d| Instant::now() + Duration::from(d));
-
         // setup control-c handling
-        let running = Arc::new(AtomicBool::new(true));
-        let r = running.clone();
+        let workload_runner_shutdown_signaler = workload_runner.shutdown_signaler();
```
peterschwarz, line 172: A better way to do this is to use a simple channel, in the ctrlc handler and then just shutdown the workload runner, without the extra signaller. For example:

```rust
        let (shutdown_tx, shutdown_rx) = std::sync:mpsc::channel();
        ctrlc::set_handler(move || {
            if let Err(err) = shutdown_tx.send(()) {
                // already called control-c, so process exit
            }
        })
        .map_err(|_| {
            CliError::ActionError("Unable to set up workload ctrlc handler".to_string())
        })?;

        let _ = shutdown_rx.recv().ok_or(());
        drop(shutdown_rx);
        workload_runner.shutdown(); // signal and wait, omitting error handling for breviety 
```

isabeltomb: I think this would create a similar issue as before where if there is an error in the loop in `fn build()` it will break the loop and stop submitting workloads but the program will stall on this line `let _ = shutdown_rx.recv().ok_or(());` and not exit until a ctrl-c is received.

peterschwarz (approved):

PR #251, opened 2021-11-04
`family-command-workload` stabilization updates

isabeltomb:
Stabilization updates for the `family-command-workload` feature
- add a module doc comment for the command::workload module


agunde406 (approved):

vaporos (changes requested):
family-command-test doesn't feel like a good feature as "-test" is about how its currently used, not what it is.

vaporos (approved):

agunde406 (approved):

PR #276, opened 2021-11-22
`workload` stabilization updates - Add `workload-smallbank` feature

isabeltomb:
Add a "workload-smallbank" feature to transact-cli and put any code
currently behind the "workload" feature, that is related to smallbank,
behind this new feature.

Signed-off-by: Isabel Tomb <tomb@bitwise.io>

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -60,12 +60,14 @@ experimental = [
     # The following features are experimental:
     "command",
     "playlist",
-    "workload"
+    "workload",
+    "workload-smallbank"
 ]
 
 command = ["cylinder", "hex", "protobuf", "reqwest", "serde", "serde_json"]
 playlist = ["cylinder", "serde_json"]
-workload = ["ctrlc", "cylinder", "rand", "transact/family-command-workload", "transact/family-smallbank-workload"]
+workload = ["ctrlc", "cylinder", "rand", "transact/family-command-workload"]
+workload-smallbank = ["workload", "transact/family-smallbank-workload"]
 
 [package.metadata.deb]
 maintainer = "The Transact Team"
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -60,76 +60,90 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
         (@arg quiet: -q --quiet +global "Do not display output")
         (@setting SubcommandRequiredElseHelp));
 
-    #[cfg(feature = "workload")]
+    #[cfg(any(feature = "workload", feature = "workload-smallbank"))]
```
peterschwarz, line 63: Should this just be on workload? (given that "workload-smallbank" depends on "workload"?

cli/Cargo.toml
```diff
@@ -60,12 +60,14 @@ experimental = [
...
     "playlist",
-    "workload"
+    "workload",
+    "workload-smallbank"
 ]
 
 command = ["cylinder", "hex", "protobuf", "reqwest", "serde", "serde_json"]
 playlist = ["cylinder", "serde_json"]
-workload = ["ctrlc", "cylinder", "rand", "transact/family-command-workload", "transact/family-smallbank-workload"]
+workload = ["ctrlc", "cylinder", "rand", "transact/family-command-workload"]
```
peterschwarz, line 69: Should there also be a "workload-command" feature? Otherwise, getting smallbank also gets you command family

cli/src/main.rs
isabeltomb: Fixed the feature guards

cli/Cargo.toml
isabeltomb: Based on discussion - Leaving as is with the intention of removing the "workload-smallbank" feature after it's stabilized

peterschwarz (approved):

agunde406 (approved):

PR #282, opened 2021-11-22
Update rand dependency

vaporos:
Removes rand_hc and updates rand to 0.8.

### DIFF
--- a/libtransact/src/families/xo.rs
+++ b/libtransact/src/families/xo.rs
@@ -43,9 +43,9 @@ impl TransactionWorkload for XoTransactionWorkload {
     fn next_transaction(
         &mut self,
     ) -> Result<(TransactionPair, Option<ExpectedBatchResult>), InvalidStateError> {
-        let nonce = self
-            .rng
-            .sample_iter(&Alphanumeric)
+        let nonce = std::iter::repeat(())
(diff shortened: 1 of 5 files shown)


libtransact/src/families/xo.rs
```diff
@@ -43,9 +43,9 @@ impl TransactionWorkload for XoTransactionWorkload {
     fn next_transaction(
         &mut self,
     ) -> Result<(TransactionPair, Option<ExpectedBatchResult>), InvalidStateError> {
-        let nonce = self
-            .rng
-            .sample_iter(&Alphanumeric)
+        let nonce = std::iter::repeat(())
+            .map(|()| self.rng.sample(Alphanumeric))
```
peterschwarz, line 47: Should this just continue to use `sample_iter`, as it does in `protocol/transaction.rs`

vaporos: As far as I can tell, this is the suggested way to do this; as previously written with sample_iter(), this code would no longer compile with this version of rand. The difference is that this code is using self.rng (because we seed it) and the protocol/transaction.rs code is using rand::thread_rng(), so the protocol/transaction.rs code is less sensitive to ownership issues.

agunde406 (approved):

peterschwarz (approved):

PR #284, opened 2021-11-23
`workload-batch-gen` stabilization updates - fix protobuf use

isabeltomb:
Stabilization updates for the libtransact `workload-batch-gen` feature, this PR limits the use of Protobuf in the `workload-batch-gen` feature.

- Add two traits `BatchReader` and `TransactionReader`, which provide an API for reading batches and reading transactions from a source.
- Replace the `LengthDelimitedMessageSource` with `ProtobufTransactionReader` and `ProtobufBatchReader` which implement`TransactionReader` and `BatchReader`.
- Update tests to use `ProtobufTransactionReader` and `ProtobufBatchReader`.
- Limit the direct use of protobuf in tests to only the function that writes length delimited protobuf.

### DIFF
--- a/libtransact/src/workload/batch_gen.rs
+++ b/libtransact/src/workload/batch_gen.rs
@@ -25,11 +25,11 @@ use cylinder::Signer;
 use protobuf::Message;
 
 use crate::error::{InternalError, InvalidStateError};
-use crate::protocol::batch::{BatchBuilder, BatchPair};
-use crate::protocol::transaction::Transaction as ProtocolTransaction;
+use crate::protocol::batch::{Batch, BatchBuilder, BatchPair};
+use crate::protocol::transaction::Transaction;
 use crate::protos::FromNative;
 use crate::protos::FromProto;
-use crate::protos::{batch::Batch, transaction::Transaction};
+use crate::protos::batch::Batch as ProtobufBatch;
 
 use super::error::{BatchReadingError, BatchingError};
 use super::source::LengthDelimitedMessageSource;
@@ -63,7 +63,10 @@ impl<'a> SignedBatchProducer<'a> {
         loop {
             match self.next() {
                 Some(Ok(batch)) => {
-                    if let Err(err) = batch.write_length_delimited_to_writer(writer) {
+                    let proto_batch = ProtobufBatch::from_native(batch).map_err(|err| {
+                        BatchingError::InternalError(InternalError::from_source(Box::new(err)))
+                    })?;
+                    if let Err(err) = proto_batch.write_length_delimited_to_writer(writer) {
                         return Err(BatchingError::InternalError(
                             InternalError::from_source_with_message(
                                 Box::new(err),
@@ -86,8 +89,13 @@ impl<'a> Iterator for SignedBatchProducer<'a> {
     /// Gets the next BatchResult.
     /// `Ok(None)` indicates that the underlying source has been consumed.
     fn next(&mut self) -> Option<BatchResult> {
-        let txns = match self.transaction_source.next(self.max_batch_size) {
-            Ok(txns) => txns,
+        let txns: Vec<Transaction> = match self.transaction_source.next(self.max_batch_size)
+        {
+            Ok(txns) => txns
+                .into_iter()
+                .map(Transaction::from_proto)
+                .collect::<Result<Vec<Transaction>, _>>()
+                .ok()?,
             Err(err) => {
                 return Some(Err(BatchingError::InternalError(
                     InternalError::from_source_with_message(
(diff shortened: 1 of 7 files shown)


libtransact/src/protocol/transaction.rs
```diff
@@ -64,6 +64,33 @@ pub struct TransactionHeader {
 }
 
 impl TransactionHeader {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
peterschwarz, line 68: Shouldn't there be a builder for this?

peterschwarz (changes requested):
The TransactionHeader constructor should not be added

libtransact/src/protocol/transaction.rs
isabeltomb: Definitely skipped right over the TransactionBuilder when doing this, removed this method and changed the test to use TransactionBuilder instead

peterschwarz (approved):

libtransact/src/workload/batch_gen.rs
```diff
@@ -25,7 +25,7 @@ use cylinder::Signer;
 use protobuf::Message;
 
 use crate::error::{InternalError, InvalidStateError};
-use crate::protocol::batch::{BatchBuilder, BatchPair};
+use crate::protocol::batch::{Batch as ProtocolBatch, BatchBuilder, BatchPair};
```
vaporos, line 28: I think it would be good here to use Batch/Transaction from protocol and rename the protos batch instead (ProtobufBatch?).

libtransact/src/workload/batch_gen.rs
```diff
@@ -63,7 +63,12 @@ impl<'a> SignedBatchProducer<'a> {
         loop {
             match self.next() {
                 Some(Ok(batch)) => {
-                    if let Err(err) = batch.write_length_delimited_to_writer(writer) {
+                    let proto_batch = Batch::from_native(batch).map_err(|_| {
+                        BatchingError::InvalidStateError(InvalidStateError::with_message(
+                            "Failed to build batch".into(),
```
vaporos, line 68: This should probably be an InternalError constructed with InternalError::from_source(e) where e is the error being mapped.

libtransact/src/workload/mod.rs
```diff
@@ -48,6 +48,18 @@ pub trait BatchWorkload: Send {
         -> Result<(BatchPair, Option<ExpectedBatchResult>), InvalidStateError>;
 }
 
+/// `TransactionReader` provides an API for reading transactions from a source
+pub trait TransactionReader {
+    /// Returns a vec with at least max_txns `Transactions` in it
+    fn next(&mut self, max_txns: usize) -> Result<Vec<Transaction>, InvalidStateError>;
```
vaporos, line 54: These should return InternalError. InvalidStateError is primarily for things like builders that the user can get into an invalid state by calling inconsistent functions or insufficient functions prior to build().

libtransact/src/workload/mod.rs
```diff
@@ -48,6 +48,18 @@ pub trait BatchWorkload: Send {
...
+}
+
+/// `BatchReader` provides an API for reading batches from a source
+pub trait BatchReader {
+    /// Returns a vec with at least max_batches `Batches` in it
+    fn next(&mut self, max_batches: usize) -> Result<Vec<BatchPair>, InvalidStateError>;
```
vaporos, line 60: Same here, InternalError.

libtransact/src/workload/batch_gen.rs
```diff
@@ -26,15 +26,15 @@ use protobuf::Message;
 
 use crate::error::{InternalError, InvalidStateError};
 use crate::protocol::batch::{Batch as ProtocolBatch, BatchBuilder, BatchPair};
-use crate::protocol::transaction::Transaction as ProtocolTransaction;
+use crate::protocol::transaction::Transaction;
+use crate::protos::batch::Batch;
```
vaporos, line 30: We should maybe call this ProtobufBatch here too since we have the protocol Transaction and usually Batch/Transaction work together.

libtransact/src/workload/source.rs
```diff
@@ -16,48 +16,111 @@
...
-/// Decodes Protocol Buffer messages from a length-delimited input reader.
-pub struct LengthDelimitedMessageSource<'a, T: 'a> {
+use crate::error::InvalidStateError;
+use crate::protocol::batch::BatchPair;
+use crate::protocol::transaction::Transaction;
+use crate::protos::FromProto;
+use crate::workload::{BatchReader, TransactionReader};
+
+/// Decodes Protocol Buffer `Batches` from a length-delimited input reader.
+pub struct ProtobufBatchReader<'a> {
```
vaporos, line 32: Put ProtobufBatchReader/ProtobufBatchWriter in a separate module (or two) so it cleans up the imports (it should be clear that only that code pulls in anything from protos.

libtransact/src/workload/batch_gen.rs
```diff
@@ -162,17 +162,19 @@ mod tests {
 
     use cylinder::{secp256k1::Secp256k1Context, Context, Signature, Signer};
 
-    use crate::protos::batch::BatchHeader;
-    use crate::protos::transaction::TransactionHeader;
+    use crate::protocol::batch::BatchHeader;
+    use crate::protocol::transaction::{HashMethod, TransactionBuilder};
+    use crate::protos::transaction::Transaction as ProtoTransaction;
```
vaporos, line 167: Call this ProtobufTransaction if you take my other suggestions for ProtocolBatch naming, etc.

libtransact/src/workload/batch_gen.rs
isabeltomb: Fixed these to use protocol Batch/Transaction and rename protos Batch/Transaction to ProtobufBatch/ProtobufTransaction

libtransact/src/workload/batch_gen.rs
isabeltomb: Fixed

libtransact/src/workload/mod.rs
isabeltomb: Fixed

libtransact/src/workload/mod.rs
isabeltomb: Fixed

libtransact/src/workload/batch_gen.rs
isabeltomb: Fixed

libtransact/src/workload/source.rs
isabeltomb: The traits are now in their own `batch_reader` and `transaction_reader` modules each with their own `protobuf` sub-module containing the protobuf impls

libtransact/src/workload/batch_gen.rs
isabeltomb: Fixed

vaporos (approved):

peterschwarz (approved):

PR #286, opened 2021-11-23
Drop contexts in serial sched when txns complete

peterschwarz:
This change drops the appropriate contexts when a transaction is complete or the schedule is canceled.

It updates the current transaction to include its associated context ID so that the context can be dropped on cancellation.  Otherwise, the context would get leaked.

### DIFF
--- a/libtransact/src/scheduler/serial/core.rs
+++ b/libtransact/src/scheduler/serial/core.rs
@@ -111,6 +111,12 @@ impl From<std::sync::PoisonError<std::sync::MutexGuard<'_, Shared>>> for CoreErr
     }
 }
 
+impl From<crate::error::InternalError> for CoreError {
+    fn from(error: crate::error::InternalError) -> Self {
+        CoreError::Internal(error.to_string())
+    }
+}
+
 pub struct SchedulerCore {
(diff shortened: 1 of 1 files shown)


libtransact/src/scheduler/serial/core.rs
```diff
@@ -407,8 +429,11 @@ impl SchedulerCore {
                     sender.send(self.current_batch.take()).map_err(|_| {
                         CoreError::Internal("aborted batch receiver dropped".into())
                     })?;
-                    // Also remove the current transaction
-                    self.current_txn.take();
+
+                    // Also remove the drop the current transaction's context, if it exists
```
agunde406, line 433: Typo in comment: Also remove the drop the

agunde406 (changes requested):

vaporos (approved):

libtransact/src/scheduler/serial/core.rs
peterschwarz: Fixed

agunde406 (approved):

vaporos (approved):

PR #296, opened 2021-12-01
Move `CommandTransactionBuilder`

isabeltomb:
Move the `CommandTransactionBuilder`, from the `family::command::workload` module, to a new `family::command::transaction_builder` module behind a new feature "family-command-transaction-builder" and re-export it from `family::command`. Additionally, depreciate the builder in the workload module.

### DIFF
--- a/libtransact/Cargo.toml
+++ b/libtransact/Cargo.toml
@@ -84,6 +84,7 @@ stable = [
     # The following features are stable:
     "contract-archive",
     "family-command",
+    "family-command-transaction-builder",
     "family-command-workload",
     "family-smallbank",
     "postgres",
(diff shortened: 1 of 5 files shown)


vaporos (changes requested):
We should put this in a new feature probably so it is possible to pull in the handler without the extra dependencies (cylinder for example).

isabeltomb:
> We should put this in a new feature probably so it is possible to pull in the handler without the extra dependencies (cylinder for example).

Put this behind a new feature "family-command-transaction-builder"

peterschwarz:
the workload module should `pub use` the builder for backwards compatibility - otherwise it is a API-breaking change.  I think we can deprecate it there, too

isabeltomb:
> the workload module should `pub use` the builder for backwards compatibility - otherwise it is a API-breaking change. I think we can deprecate it there, too

Deprecating a `pub use` apparently doesn't work so I left the original `CommandTransactionBuilder` in the workload module and deprecated it there

libtransact/Cargo.toml
```diff
@@ -137,13 +138,8 @@ contract-context-key-value = ["contract-context", "key-value-state"]
...
 execution = ["context", "handler", "log", "protocol-transaction", "scheduler"]
 family-command = ["handler"]
-family-command-workload = [
-    "cylinder",
-    "family-command",
-    "protocol-sabre",
-    "protocol-transaction-builder",
-    "workload",
-]
+family-command-transaction-builder = ["cylinder", "protocol-transaction-builder"]
```
vaporos, line 141: As a sub-module, this should probably pull in family-command as well?

isabeltomb: Added "family-command"

peterschwarz (approved):

vaporos (approved):

PR #297, opened 2021-12-02
Fix clippy errors

isabeltomb:


libtransact/src/workload/runner.rs
```diff
@@ -1002,9 +1002,9 @@ pub struct Link {
 /// Used for deserializing `GET /batch_status` responses.
 #[derive(Debug, Deserialize)]
 struct BatchInfo {
-    pub id: String,
+    pub _id: String,
```
peterschwarz, line 1005: If these fields are not used, we should remove them, but add a comment with an example of the complete response, as of this commit.  

E.g.
```
/// This struct is a subset of the batch status response.  The complete response looks like
/// {
///     id: "<id string>",
```
etc

This applies to the child structs/enums of this object, as well.

peterschwarz (changes requested):

libtransact/src/workload/runner.rs
isabeltomb: Did this for `BatchInfo` but `InvalidTransaction` and `ValidTransaction` are printed out in error messages if a batch result doesn't match the expected result, so removing the "unused fields" would remove some important information from those error messages. Implemented display for `InvalidTransaction` and `ValidTransaction` instead.

peterschwarz: Discussing this further, the right fix here would be a) remove the `pub` modifiers from the fields, and then add `#[allow(dead_code)]` to each with a comment of "these are deserialized and available for debugging"

isabeltomb: Fixed

peterschwarz (approved):

agunde406 (approved):

PR #303, opened 2022-01-06
Alter column type for leaf address

peterschwarz:
Prior to this change, the column type for merkle_radix_leaf.address was specified as `"STRING"`. As this type is not one of the types automatically treated as `TEXT`, it defaults to being treated as a `NUMERIC` value.  With certain addresses, where all characters were valid decimal digits with a leading zero, the leading zero would be dropped. This would create
instances where the address value would not be returned, but not with the same address queried.

Changing the column type to `"TEXT"` fixes the problem.

The migration added will also correct all address that are missing a leading zero.

This change also adds several tests for both SQLite and Postgres to guard against this issue in the future, while also exercising the `list_leaves` trait function implementations of each.

### DIFF
--- a/libtransact/src/state/merkle/sql/migration/sqlite/migrations/2022-01-06-101500-correct-column-type/down.sql
+++ b/libtransact/src/state/merkle/sql/migration/sqlite/migrations/2022-01-06-101500-correct-column-type/down.sql
@@ -1,0 +1,35 @@
+-- Copyright 2022 Cargill Incorporated
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


libtransact/src/state/merkle/sql/migration/sqlite/migrations/2022-01-06-101500-correct-column-type/down.sql
```diff
@@ -0,0 +1,35 @@
+-- Copyright 2021 Cargill Incorporated
```
agunde406, line 1: This should be 2022, same for the following file

agunde406 (approved):

vaporos (approved):

PR #368, opened 2022-07-21
Fix midpoint pruning

peterschwarz:
This PR updates the changelog after pruning by taking into account pruning occurring in the middle of a chain of state roots.  The changelog is updated such that the parent and successor nodes are properly re-linked with the removal of the intermediate node.  It also limits the pruning of changelog additions to those candidates that were deleted.

### DIFF
--- a/libtransact/src/state/error.rs
+++ b/libtransact/src/state/error.rs
@@ -141,3 +141,30 @@ impl From<InvalidStateError> for StateError {
         Self::InvalidState(err)
     }
 }
+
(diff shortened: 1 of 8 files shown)


agunde406 (approved):

rbuysse:
I tested this PR locally with the postgres migration test in https://github.com/Cargill/splinter/pull/2031 and the test no longer reports merkle_radix foreign key constraint errors.

isabeltomb (approved):

